PR #13, opened 2017-11-16
Analyzer Runner Tool

ivanbasov:
This is a copy of https://github.com/dotnet/roslyn/pull/23087
I copied props and target files from roslyn. I think there can be a lot of unused tag there. Please let me know if we prefer to clean them or to keep as close to roslyn as possible.


ivanbasov:
Decided to keep it in roslyn. The right PR is https://github.com/dotnet/roslyn/pull/23087

PR #28, opened 2017-12-15
Change project names to be consistent with assembly names.

333fred:
RepoToolset has some assumptions about project name == assembly name when including things for XAML files. This updates all the projects in the sdk solution to use their assembly name as their project name.


jmarolf:
@333fred first pass looks good.  I'll verify the setup authoring.

jmarolf:
Looks good

PR #84, opened 2017-12-21
Syntax Visualizer accessibility - selected tree item in Syntax Tree is not visible in HC white theme

chborl:
Fixes bug https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems?id=483440 

In HC White mode, when selecting an item in the Syntax Tree pane of the Syntax Visualizer, both the text and the background were dark so the text could not be read.

### DIFF
--- a/src/Tools/SyntaxVisualizer/Roslyn.SyntaxVisualizer.Control/SyntaxVisualizerControl.xaml.cs
+++ b/src/Tools/SyntaxVisualizer/Roslyn.SyntaxVisualizer.Control/SyntaxVisualizerControl.xaml.cs
@@ -35,6 +35,8 @@ namespace Roslyn.SyntaxVisualizer.Control
 
         #region Private State
         private TreeViewItem _currentSelection;
+        private Brush _previousForeground;
+        private Brush _previousBackground;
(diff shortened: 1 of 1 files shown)


chborl:
@dotnet/roslyn-analysis  @dotnet/roslyn-ide  for review

src/Tools/SyntaxVisualizer/Roslyn.SyntaxVisualizer.Control/SyntaxVisualizerControl.xaml.cs
```diff
@@ -35,6 +35,8 @@ private class SyntaxTag
 
         #region Private State
         private TreeViewItem _currentSelection;
+        private Brush _previousForeground;
```
jasonmalinowski: This seems fishy that we're having to muck around with this in the first place. My gut is we're missing some XAML declaration somewhere that can do this declaratively.

What's the root cause of this? Why is this the best approach?

dpoeschl:
Superseded by https://github.com/dotnet/roslyn-sdk/pull/90

PR #90, opened 2018-01-28
Syntax Visualizer treeview accessible colors for selected item

dpoeschl:
Fixes bug https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems?id=483440

(This also changes the semantics of `_currentSelection` when there is no selection. Previously, it remembered the last selection, even when the treeview containing that selection went away (i.e. closing all documents). Now, it nulls out and we watch for nulls.)

### Customer scenario

Sight impaired users trying to use the Syntax Visualizer may have a difficult time reading the currently selected item's text in both regular and high contrast themes.

### Bugs this fixes

https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems?id=483440

### Workarounds, if any

Unselect the given item to read its text and then re-select it

### Risk

Risk is low. Nothing builds on top of this feature, and this change mostly just sets colors and handles nulls more coherently than before.

### Performance impact

Extremely low. We do a tiny bit more work each time an item in the treeview is selected, but that happens once per user action and not in any type of loop.

### Is this a regression from a previous update?

No.

### Root cause analysis

Accessibility was not considered when the Syntax Visualizer was created.

### How was the bug found?

Accessibility testers


dpoeschl:
@chborl I've included the commit from your PR here, and built upon it. The real key here was [this one line](https://github.com/dotnet/roslyn-sdk/pull/90/files#diff-4192aa7778be11de6ac206dc3b0d6410R71).

dpoeschl:
Tagging for review: @dotnet/roslyn-analysis @dotnet/roslyn-ide

dpoeschl:
FYI @matthejo

jinujoseph:
{ Pending shiproom approval}

jinujoseph:
Approved to merge via [link](https://devdiv.visualstudio.com/DevDiv/NET%20Developer%20Experience%20Productivity/_workitems/edit/483440)

PR #92, opened 2018-02-06
updating repo toolset

jmarolf:
infrastructure only change.


sharwell:
@jmarolf does this fix the tmp_proj bug or should I follow up on that separately?

jmarolf:
@sharwell I do not believe this solves that. We'll need to add workarounds to https://github.com/dotnet/roslyn-tools/tree/master/sdks/RepoToolset

PR #93, opened 2018-02-07
Prevent deployment of the VS 2015 SDK in VS 2017+

sharwell:
This was a contributor pain point, and led to incorrect loading of binaries in the IDE.


jmarolf:
@sharwell I am fine with this change, but can you file a bug on this repo that we need to have a story for people on VS 2015 who want to contribute to the VS 2015 extension?

sharwell:
@jmarolf VS 2015 can't load, deploy, or debug the extension. For all practical purposes the code is an orphan at this point.

sharwell:
➡️ Filed #96

Issue #96, opened 2018-02-07
Remove or fix Visual Studio 2015 code


sharwell:
Currently the repository contains an extension for Visual Studio 2015, but there is no way for contributors to load, deploy, or debug the SDK in this target environment.

The contributor experience for the Visual Studio 2015 SDK should be fixed and documented, or the code should be removed from this repository.

sharwell:
The extension has since been removed.

PR #100, opened 2018-02-08
Use VS 2015 dependencies

jmarolf:
fixes https://github.com/dotnet/roslyn-sdk/issues/96

### DIFF
--- a/build/Versions.props
+++ b/build/Versions.props
@@ -34,21 +34,21 @@
     <VSSDKTemplateWizardInterfaceVersion>12.0.4</VSSDKTemplateWizardInterfaceVersion>
     <EnvDTEVersion>8.0.1</EnvDTEVersion>
     <EnvDTE80Version>8.0.0</EnvDTE80Version>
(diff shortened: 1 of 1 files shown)


jasonmalinowski:
@jmarolf Should this be commented somewhere in that file so people don't try to "helpfully" upgrade?

sharwell:
@jmarolf This does not fix the support for 2015. Visual Studio 2015 does not support the new project system, and the VSSDK does not support cross-deploying from a different version of VS.

jmarolf:
@sharwell 

> Visual Studio 2015 does not support the new project system

Correct which is why the VS2015 extension uses the PCL templates under [src/Templates/VS2015](https://github.com/dotnet/roslyn-sdk/tree/master/src/Templates/VS2015)

If you build without this change `artifacts\Debug\VSSetup\Roslyn.SDK.VS2015.vsix` will produce templates that work on VS 2015, but the assemblies we build with prevent the syntax visualizer from loading on VS 2015

PR #118, opened 2018-04-11
Update for new MSBuildWorkspace

DustinCampbell:
This change updates existing sample that use `MSBuildWorkspace` to the new `Microsoft.CodeAnalysis.Workspaces.MSBuild` NuGet package, and adds a new "SolutionExplorer" sample that significantly exercises `MSBuildWorkspace`.

Because `Microsoft.CodeAnalysis.Workspaces.MSBuild` is not released on https://nuget.org yet, I've added a temporary `NuGet.config` that includes the roslyn MyGet feed. In addition, the package versions used are 2.9.0 beta releases from that feed. Eventually, the `NuGet.config` will go away and the package versions will be updated to stable versions.


samples/CSharp/FormatSolution/FormatSolution.csproj
```diff
@@ -14,9 +14,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Build" Version="15.3.409" />
-    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="15.3.409" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.6.0" />
+    <PackageReference Include="Microsoft.Build.Locator" Version="1.0.13" />
+
+    <!-- These package versions should be updated to 2.9.0 when released on NuGet -->
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="2.9.0-beta4-62810-05" />
```
Pilchie: Hmm - it looks like we screwed up the versioning of 2.9 already.  We probably don't want these packages to have `-beta4` in them.

@jaredpar @jasonmalinowski ?

DustinCampbell: I have no idea. Those are the packages on the roslyn feed.

jasonmalinowski: Ah, cute. Uh, I guess we could bump this back to -beta1 if we wanted to. @brettfo?

DustinCampbell: what if we need beta4 in 15.8?

jmarolf:
Could we merge these into https://github.com/dotnet/roslyn-sdk/tree/dev15.8.x? I would like samples in master to be against public nuget packages so people can reuse the samples without adding a new feed.

DustinCampbell:
@jmarolf: Sure! I didn't realize we had separate branches for future work.

PR #131, opened 2018-08-14
Add diagnostic and code fix verification helpers

sharwell:
* General cleanup
    * [x] Better documentation (DotNetAnalyzers/StyleCopAnalyzers#279 and more)
    * [x] Asynchronous methods (DotNetAnalyzers/StyleCopAnalyzers#279, DotNetAnalyzers/StyleCopAnalyzers#550)
    * [x] De-duplication of diagnostics (driver used to call some methods too many times) (DotNetAnalyzers/StyleCopAnalyzers@cc9a15a)
    * [x] Disallow undeclared compiler errors (DotNetAnalyzers/StyleCopAnalyzers@746e551)
    * [x] Compile as DLL, allow unsafe (DotNetAnalyzers/StyleCopAnalyzers@4ee468e)
    * [x] Make sure analyzers are enabled (DotNetAnalyzers/StyleCopAnalyzers#841, DotNetAnalyzers/StyleCopAnalyzers#951)
    * [x] Treat AD0001 as a test failure (DotNetAnalyzers/StyleCopAnalyzers#889)
    * [x] Cleanup to adhere to new API (DotNetAnalyzers/StyleCopAnalyzers#926)
    * [x] Enable documentation comments (DotNetAnalyzers/StyleCopAnalyzers#1055)
    * [x] Accuracy of TestDiagnosticProvider? (DotNetAnalyzers/StyleCopAnalyzers#1279)
    * [x] Deterministic diagnostic order (DotNetAnalyzers/StyleCopAnalyzers#1741)
    * [x] EmptyDiagnosticResults (DotNetAnalyzers/StyleCopAnalyzers#412)
    * [x] Better handling of code fix validation (DotNetAnalyzers/StyleCopAnalyzers#447)
    * [x] Switch to xUnit (DotNetAnalyzers/StyleCopAnalyzers#517)
    * [x] Overload VerifyCSharpDiagnosticAsync for single diagnostic case (DotNetAnalyzers/StyleCopAnalyzers#526)
    * [x] Improved code fix testing (DotNetAnalyzers/StyleCopAnalyzers#935, DotNetAnalyzers/StyleCopAnalyzers#1203)
    * [x] ~~GetOfferedCSharpFixesAsync (DotNetAnalyzers/StyleCopAnalyzers#1250)~~ (likely not required now that iteration limit properties support the value `0`)
    * [x] Require fixer for code fix testing (DotNetAnalyzers/StyleCopAnalyzers#1299)
[...]

PR #137, opened 2018-08-20
[WIP] Inital work on Microsoft.CodeAnalysis.Testing

jmarolf:
Liberally stealing from @sharwell's work, here is an initial set of nuget packages

Ideal usecase for this API is to reference the version for your test framework (MSTest, xUnit. NUnit) and it just works:tm:!
```C#
using System.Threading.Tasks;
using AnalyzerNamespace;
using Xunit;
using Verifier = Microsoft.CodeAnalysis.Testing.xUnit.CodefixVerifier<SampleAnalyzer, SampleCodeFix>;

namespace Analyzer.Tests
{
    public class ExampleTestClass
    {
        [Fact]
        public async Task ExampleTest()
        {
            var test = @"
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using System.Diagnostics;

namespace ConsoleApplication1
{
    class [|TypeName|]
    {   
    }
}";
            await Verifier.VerifyDiagnosticAsync(test);
        }
    }
}
```

### TODO
- [X] Simplify namespaces and type names
- [ ] Rationalize Metadata API for consumer
- [X] Rationalize version of roslyn vs. versions of the test framework
- [ ] Add support for Refactoring Verifiers
- [ ] Migrate all roslyn-sdk tests to use new framework
- [ ] Update templates to use new framework

**CC:** @sharwell @DustinCampbell @kuhlenh @mavasani


src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
```diff
@@ -0,0 +1,21 @@
...
+  
+  <PropertyGroup>
+    <IsPackable>true</IsPackable>
+    <Description>Roslyn Analyzer Test Framework Common Types.</Description>
+    <PackageReleaseNotes>Roslyn Analyzer Test Framework Common Types</PackageReleaseNotes>
+    <PackageTags>Roslyn Analyzer Test Framework Common</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="2.9.0" />
```
sharwell: :exclamation: needs to be 1.2.1.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/BaseAnalyzerTest.cs
```diff
@@ -0,0 +1,800 @@
...
+            var modifiedSpecificDiagnosticOptions = supportedDiagnosticsSpecificOptions.ToImmutableDictionary().SetItems(project.CompilationOptions.SpecificDiagnosticOptions);
+            var modifiedCompilationOptions = project.CompilationOptions.WithSpecificDiagnosticOptions(modifiedSpecificDiagnosticOptions);
+
+            var solution = project.Solution.WithProjectCompilationOptions(project.Id, modifiedCompilationOptions);
+            return solution.GetProject(project.Id);
+        }
+
+        public virtual AdhocWorkspace CreateWorkspace()
+        {
+            return new AdhocWorkspace();
```
sharwell: :exclamation: this must use vs-threading for correctness.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CSharp.Analyzer.Testing/CSharpDiagnosticVerifier.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.CodeAnalysis.CSharp.Analyzer.Testing
+{
+    public class CSharpDiagnosticVerifier<TAnalyzer, TVerifier> : DiagnosticVerifier<TAnalyzer, CSharpAnalyzerTest<TAnalyzer, TVerifier>, TVerifier>
+        where TAnalyzer : DiagnosticAnalyzer, new()
+        where TVerifier : IVerifier, new()
+    {
+    }
+
+    public static class CSharpDiagnosticVerifier<TAnalyzer>
```
sharwell: :bulb: One file per top-level type to ensure, among other things, <kbd>t</kbd> operates as Navigate To Type on GitHub.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.VisualBasic.Codefix.Testing/VisualBasicCodefixVerifier.vb
```diff
@@ -0,0 +1,12 @@
+﻿Imports Microsoft.CodeAnalysis.Analyzer.Testing
+Imports Microsoft.CodeAnalysis.Codefix.Testing
+Imports Microsoft.CodeAnalysis.CodeFixes
+Imports Microsoft.CodeAnalysis.Diagnostics
+
+Public Class VisualBasicCodefixVerifier(Of TAnalyzer As {DiagnosticAnalyzer, New}, TCodeFix As {CodeFixProvider, New}, TVerifier As {IVerifier, New})
+    Inherits CodeFixVerifier(Of TAnalyzer, TCodeFix, VisualBasicCodeFixTest(Of TAnalyzer, TCodeFix, TVerifier), TVerifier)
+End Class
+
+Public Class DefaultVisualBasicCodefixVerifier(Of TAnalyzer As {DiagnosticAnalyzer, New}, TCodeFix As {CodeFixProvider, New})
```
sharwell: :bulb: One file per top-level type

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
```diff
@@ -0,0 +1,589 @@
...
+                    // Allow the operations to run in parallel
+                    await t1;
+                    await t2;
+                    await t3;
+                    await t4;
+                }
+            }
+        }
+
+        private async Task VerifyFixpublicAsync(
```
sharwell: 😕 What is `Fixpublic`?

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/CodeFixVerifier.cs
```diff
@@ -0,0 +1,45 @@
...
+        public static DiagnosticResult CompilerError(string errorIdentifier)
+            => DiagnosticVerifier<TAnalyzer, TTest, TVerifier>.CompilerError(errorIdentifier);
+
+        public static Task VerifyDiagnosticAsync(string source, DiagnosticResult expected, CancellationToken cancellationToken = default)
+            => DiagnosticVerifier<TAnalyzer, TTest, TVerifier>.VerifyDiagnosticAsync(source, expected, cancellationToken);
+
+        public static Task VerifyFixAsync(string source, DiagnosticResult expected, string fixedSource, CancellationToken cancellationToken = default)
+            => VerifyixAsync(source, new[] { expected }, fixedSource, cancellationToken);
+
+        public static Task VerifyixAsync(string source, DiagnosticResult[] expected, string fixedSource, CancellationToken cancellationToken = default)
```
sharwell: ❕ `VerifyFixAsync`

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/CodeFixVerifier.cs
```diff
@@ -0,0 +1,45 @@
...
+        public static DiagnosticResult Diagnostic(string diagnosticId = null)
+            => DiagnosticVerifier<TAnalyzer, TTest, TVerifier>.Diagnostic(diagnosticId);
+
+        public static DiagnosticResult Diagnostic(DiagnosticDescriptor descriptor)
+            => DiagnosticVerifier<TAnalyzer, TTest, TVerifier>.Diagnostic(descriptor);
+
+        public static DiagnosticResult CompilerError(string errorIdentifier)
+            => DiagnosticVerifier<TAnalyzer, TTest, TVerifier>.CompilerError(errorIdentifier);
+
+        public static Task VerifyDiagnosticAsync(string source, DiagnosticResult expected, CancellationToken cancellationToken = default)
```
sharwell: ❓ Missing overload that takes `DiagnosticResult[]`?

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/CodeFixVerifier.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Analyzer.Testing;
+
+namespace Microsoft.CodeAnalysis.Codefix.Testing
+{
+    public class CodeFixVerifier<TAnalyzer, TCodeFix, TTest, TVerifier>
```
sharwell: ❕ These types all need to be `static`. While this isn't a hard requirement for using the test framework with `using Verifier = ...` syntax, the alternative approach using `using static` doesn't consider static methods defined in base types, so allowing inheritance in these verifier types only leads to confusion.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
mavasani: Do we need to also release a package based on 2.6.0 to enable testing IOperation analyzers?

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
```diff
@@ -0,0 +1,589 @@
...
+                {
+                    throw new Exception($"Expected '{expectedNumberOfIterations}' iterations but found '{expectedNumberOfIterations - numberOfIterations}' iterations.");
+                }
+            }
+
+            return project;
+        }
+
+        /// <summary>
+        /// Implements a workaround for issue #936, force re-parsing to get the same sort of syntax tree as the original document.
```
sharwell: :bulb: Could expand this to a full link to https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/936

mavasani: Please remove this workaround - this basically hides bugs in Roslyn. See https://github.com/dotnet/roslyn-analyzers/pull/1525 and discussion in it. Majority of Roslyn team prefers not to take this approach.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
jmarolf: @sharwell 1.0.1 is the oldest version for which nuget dependencies are correct (1.0.0 got this wrong)

@mavasani This points out an important problem.  We could release mirrored packages such that, for whatever version of roslyn the user is targeting, an equivalent testing library exists.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
jmarolf: :+1:

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/CodeFixVerifier.cs
jmarolf: 👍 I'll fix it

jmarolf:
@sharwell 

> Did you keep track of all the changes you made relative to #131?

I'll write up a diff.  The hope is that it's mostly the same except for being test platform agnostic on the lover layers

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
jmarolf: @mavasani what other analyzer features have we shipped and in which versions?

| Version | Feature             |
| :-----: | :-----------------: |
| 2.0?    | Concurrent analysis |
| 2.6     | IOperation          |

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
```diff
@@ -0,0 +1,589 @@
...
+
+                FixAllContext.DiagnosticProvider fixAllDiagnosticProvider = TestDiagnosticProvider.Create(analyzerDiagnostics);
+
+                var analyzerDiagnosticIds = analyzers.SelectMany(x => x.SupportedDiagnostics).Select(x => x.Id);
+                var compilerDiagnosticIds = codeFixProvider.FixableDiagnosticIds.Where(x => x.StartsWith("CS", StringComparison.Ordinal));
+                var disabledDiagnosticIds = project.CompilationOptions.SpecificDiagnosticOptions.Where(x => x.Value == ReportDiagnostic.Suppress).Select(x => x.Key);
+                var relevantIds = analyzerDiagnosticIds.Concat(compilerDiagnosticIds).Except(disabledDiagnosticIds).Distinct();
+                var fixAllContext = new FixAllContext(project.GetDocument(firstDiagnostic.Location.SourceTree), codeFixProvider, scope, equivalenceKey, relevantIds, fixAllDiagnosticProvider, cancellationToken);
+
+                var action = await fixAllProvider.GetFixAsync(fixAllContext).ConfigureAwait(false);
```
mavasani: Note that this will fail for the Well-known [batch fixer](http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/CodeFixes/FixAllOccurrences/WellKnownFixAllProviders.cs,24), which is used by large number of code fixers, if the test framework is targeting Microsoft.CodeAnalysis 2.6.0 or lower. This was fixed with https://github.com/dotnet/roslyn/pull/23492, and we recently moved roslyn-analyzers repo to 2.9 bits and now the batch fixer executes fine during our code fix tests.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
sharwell: > Concurrent analysis

1.2.x

> IOperation

2.x, but the test framework doesn't need to know about it. StyleCop Analyzers uses assembly binding redirection to test things that require 2.x even though the core of the test harness references 1.2.1.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
sharwell: ❗️ We may need a central hook for this for StyleCop Analyzers. I will experiment to see if removing it would be a hard adoption blocker.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
sharwell: I would be fine implementing a workaround (MEF-exporting an implementation of the required interface from the test harness itself, as part of `CreateSolution`) or simply documenting it as a known issue.

mavasani:
@jmarolf I haven't been able to take a thorough look at the changes yet, but a quick question: Does the test framework support testing analyzers and code fixes that work with additional files? Quite a few of our providers in roslyn-analyzers repo are based on additional files, and the test framework supports basic testing for such providers. For example, consider the latest PR from @Pilchie: https://github.com/dotnet/roslyn-analyzers/pull/1781. We need to add this support to test framework.

sharwell:
> Does the test framework support testing analyzers and code fixes that work with additional files?

Yes. It also supports analyzers that need to resolve `<include>` elements in XML documentation, which is similar to additional files but uses a different API.

mavasani:
@sharwell Thanks. Can you point me to the Verify API that takes source + additional files?

mavasani:
@jmarolf A good verification for completeness of the test framework could be to attempt to move some of the existing repos (roslyn-analyzers, stylecop, possibly analyzers/fixers in roslyn) to consume this package and verify there are no big adoption blockers and feature gaps that need rewriting large part of existing tests. We should probably split this work amongst the team and iterate over the supported APIs.

sharwell:
> Can you point me to the Verify API that takes source + additional files?

I'm not sure what the current form is in this pull request, but the whole thing originated from DotNetAnalyzers/StyleCopAnalyzers#2751 which uses this:

```csharp
await new CSharpTest
{
  TestCode = testCode,
  SolutionTransforms =
  {
    (solution, projectId) =>
    {
      return solution.AddAdditionalDocument(DocumentId.CreateNewId(projectId), additionalFileName, additionalFileContent);
    },
  },
}
```

This is a bit inconvenient, but it also predated the introduction of `SourceFileList`. At this point, I would expect the next iteration of this library to add an `AdditionalFiles` property or similar that is backed by either `SourceFileList` (for file name inference) or `List<(string fileName, string content)>` (for explicit names only).

I'm trying to avoid the introduction of a large number of "Verify" APIs, or APIs with optional parameters, in favor of focusing on the core use cases and allowing the object initializer syntax to support a very wide range of additional cases. I believe the approach offers improved long-term versioning support and reduces the chances of a user incorrectly passing parameters in situations where order matters but names do not.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
sharwell: @mavasani I reviewed the use of this method today, and it is essential for correct behavior of the test framework. The code fix verifier runs multiple iterations where diagnostics are obtained and fixes applied, similar to what would occur if a user would experience if they applied multiple code fixes as distinct operations in sequence. The editor applies this normalization step automatically; the test does not inherit this functionality automatically and must manually ensure the state is correct at the beginning of each iteration.

This code is a correctness requirement, not a convenience requirement. If it's removed, it would need to be replaced with other code that retains the correctness aspect of the original.

mavasani: > we should default the tests to an accurate representation of the behavior a user would observe

I disagree, as this hides bugs in Roslyn APIs, and we have already concluded in https://github.com/dotnet/roslyn-analyzers/pull/1525 that this workaround is undesirable by default.
Alternatively, we can include a hook that repos such as StyleCopAnalyzers can turn on to invoke this workaround to do the round tripping. As concluded in https://github.com/dotnet/roslyn-analyzers/pull/1525, the workaround should **not** be applied by _default_ as it defeats the purpose of API unit testing, one could as well write an IDE integration test if the behavior in VS IDE needs to be tested.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
mavasani: > simply documenting it as a known issue

We need to release a package based on a version of Microsoft.CodeAnalysis package that enables executing the well known batch fixer. That is the default batch fixer for our template and is used by majority of third party code fixers. Documenting it does not work is not useful as we are then not providing a solution to majority of our customers.

> StyleCop Analyzers bans the use of the well-known batch fixer outright due to its negative impact on test scenarios. Other analyzers that take a strict approach to compatibility would be likely to employ similar techniques, which this test harness helps support.

Approach taken by a specific third party analyzer package is completely irrelevant to the support being added here. We cannot add/block features from `Microsoft.CodeAnalysis.Testing` based on usage in StyleCop analyzers. If there are issues in the default Roslyn supported batch fixer, we either ought to fix it OR change the template to not advertise it. It makes no sense to have a code fix template that defaults to batch fixer and then release a test framework that explicitly does not support the batch fixer.

sharwell:
> This is really not a good solution to additional files support, which is our first class supported scenario, with large number of analyzers. I think we must include Verify APIs specific to additional files in v1 of the package.

I agree, and plan to file a follow-up PR to implement a more specific API shortly after the original PR is merged. The current support is enough to show Additional Files support is *possible* with the current API, but was omitted to help manage the size of the initial pull request.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
sharwell: > We need to release a package based on a version of Microsoft.CodeAnalysis package that enables executing the well known batch fixer.

If we can construct the MEF container in a manner that allows its use in testing older versions of Roslyn, it would certainly be an option. However, if it turns out this is not possible, it's also fine to ship a version that references older Roslyn APIs knowing that not all functionality is supported in those releases. As I mentioned previously, users who depend on newer Roslyn features with a minimum version greater than 1.0.1 will be able to apply assembly binding redirection without any negative impact on the tests. StyleCop Analyzers first introduced this strategy for churn testing after the 1.0.0 release and now depends on it as an integral part of its C# 7 test coverage.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
sharwell: > I disagree, as this hides bugs in Roslyn APIs, and we have already concluded in dotnet/roslyn-analyzers#1525 that this workaround is undesirable by default.

It's possible the usage here in this PR is a different context from the one you referenced. At minimum, the relation to the internal behavior of Roslyn APIs is substantially different. The behavior as implemented from the perspective of an external analyzer package is instrumental for building confidence in the behavior of the analyzer. I have no objection to continuing the discussion separately from this pull request, but I don't think we should deviate on this facet from the behavior proven to work in the initial iterative development.

sharwell: @mavasani How about a compromise default: include the normalization step, but by default have an assertion that the normalization step does not change the form of the tree (i.e. the normalization transform is an identity transform). This provides the guarantees of valid input to the next iteration while also validating the previous step produced an expected tree.

sharwell: The text of a compilation unit uniquely determines the syntax tree which is passed to analyzers and provided as input to code fixes. The Visual Studio IDE and the command line compiler enforce ensure this invariant is met in user scenarios. Analyzers and code fixes should be executed against this form for the optimum ability to detect incorrect behavior.

The implemention in this pull request assumes that the host is responsible for ensuring tree-based transformations produce inputs to the next diagnostic phase that adhere to the invariants for text-based input, which is consistent with the hosts mentioned above. If the host is not responsible for ensuring trees meet compiler invariants, we can remove the silent normalization. However, the invariants themselves remain. 

The normalization step ensures the analyzers are provided correct trees for analysis. The assertion in question ensures that code fixes produce the normalized form as a direct output. So we end up with one of the following:

1. Normalization is the responsibility of the host, and we silently provide the behavior while acting as a host in incremental analysis scenarios
2. Normalization is the responsibility of the code fix, and we include an assertion that the requirement is met

In practice, I believe the first case is a reasonable expectation of third party analyzer authors, for the following reasons:

1. Normalization is difficult to ensure under fix all conditions
2. Normalization is difficult to ensure for code fixes that manipulate trivia
3. It is only possible to observe a failure of the host environment to normalize the tree if a sequence of strictly tree-based transformations is applied without ever round-trip to a text-based form

The third item is an easily avoided edge case that makes the work of everyone else easier. I can understand an argument that the build-in code fixes should produce normalized form outputs, but it's not a reasonable expectation for third party analyzer authors who have enough difficulty dealing with fixes when the host *is* providing normalization. If in fact normalization is an expectation of the host, code fix tests should not fail by default when the fix both 1) produced the correct text output and 2) met the postcondition for a code fix.

sharwell: The fundamental question is:

:question: Who is responsible for ensuring the syntax tree after applying a code fix is the same as the syntax tree that *would* be produced by the compiler after parsing the text representation of the tree?

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
sharwell: ➡️ Moved to #138

sharwell:
> This is really not a good solution to additional files support, which is our first class supported scenario, with large number of analyzers. I think we must include Verify APIs specific to additional files in v1 of the package.

➡️ Filed #139 for this

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/BaseCodeFixTest.cs
mavasani: Thanks for clarifying your views. So basically, your proposal is to perform normalization to validate the syntax trees (and not just the underlying texts) coming out from a code fix are identical to the ones that will be generated after round tripping and re-parsing by the compiler? If so, that sounds a reasonable validation step. We should then change the validation methods to be as follows:

```
        private static async Task VerifyRecreateProjectDocumentsAsync(Project project, CancellationToken cancellationToken)
        {
            foreach (var documentId in project.DocumentIds)
            {
                var document = project.GetDocument(documentId);
                await VerifyRecreateDocumentAsync(document, cancellationToken).ConfigureAwait(false);
            }
             return project;
        }

        private static async Task VerifyRecreateDocumentAsync(Document document, CancellationToken cancellationToken)
        {
            var newText = await document.GetTextAsync(cancellationToken).ConfigureAwait(false);
            newText = newText.WithChanges(new TextChange(new TextSpan(0, 0), " "));
            newText = newText.WithChanges(new TextChange(new TextSpan(0, 1), string.Empty));
            var newDocument = document.WithText(newText);
            // Add code to assert that document.SyntaxTree and newDocument.SyntaxTree are identical
        }
```

sharwell: I would prefer to file a follow-up issue to add this validation primarily because the message that appears if the `Assert.Equal` in my example fails is useless (it prints the text of the trees, but we already know the text will match).

mavasani: I am fine with the above logic with couple of minor edits, which emphasize that the round tripping is an assertion of a fact that ideally **must** be respected by each code transformation provider:
1. `ValidateCodeFixTreeIsNormalized` is `true` by default.
2. `project` is updated only if `ValidateCodeFixTreeIsNormalized = false`, i.e. the non-default code path.

```diff
private static async Task<Project> RecreateProjectDocumentsAsync(Project project, CancellationToken cancellationToken)
 {
     foreach (var documentId in project.DocumentIds)
     {
         var document = project.GetDocument(documentId);
+        var initialTree = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
         document = await RecreateDocumentAsync(document, cancellationToken).ConfigureAwait(false);
+        var recreatedTree = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+        if (ValidateCodeFixTreeIsNormalized)
+        {
+            Assert.Equal(initialTree, recreatedTree);
+        }
+        else
+        {
+            project = document.Project;
+        }
     }
 
     return project;
 }
```

Forcing use of updated project with normalized documents is possible by overriding `ValidateCodeFixTreeIsNormalized = false`, but it generally indicates a bug in the code fixer and/or underlying syntax generation/update APIs used by the fixers, which will prevent such a fixer from running properly in host or application that does not normalize the output of code fixers, as this is not a requirement placed upon any host by the Roslyn syntax generation API and tool chain.

sharwell: ➡️ Moved to #140 rather than block this initial pull request

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/CodeFixVerifier.cs
```diff
@@ -23,13 +23,13 @@ public static DiagnosticResult Diagnostic(DiagnosticDescriptor descriptor)
...
+        public static Task VerifyAnalyzerAsync(string source, DiagnosticResult expected, CancellationToken cancellationToken = default)
+            => AnalyzerVerifier<TAnalyzer, TTest, TVerifier>.VerifyAnalyzerAsync(source, expected, cancellationToken);
 
-        public static Task VerifyFixAsync(string source, DiagnosticResult expected, string fixedSource, CancellationToken cancellationToken = default)
-            => VerifyixAsync(source, new[] { expected }, fixedSource, cancellationToken);
+        public static Task VerifyFCodeFixAsync(string source, DiagnosticResult expected, string fixedSource, CancellationToken cancellationToken = default)
+            => VerifyFCodeFixAsync(source, new[] { expected }, fixedSource, cancellationToken);
 
-        public static Task VerifyixAsync(string source, DiagnosticResult[] expected, string fixedSource, CancellationToken cancellationToken = default)
+        public static Task VerifyFCodeFixAsync(string source, DiagnosticResult[] expected, string fixedSource, CancellationToken cancellationToken = default)
```
sharwell: 📝 Probable typo

PR #141, opened 2018-08-24
Use Microsoft.VisualStudio.Composition when available

sharwell:
* Use vs-mef in the .NET Framework 4.5.2 target
* Will file a follow-up issue to use vs-mef in .NET Standard 1.5 build (requires adapting `ExportProvider` to `MefHostServices`)

### DIFF
--- a/src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
+++ b/src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
@@ -1,14 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.1;net452</TargetFrameworks>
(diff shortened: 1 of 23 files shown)


src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
```diff
@@ -1,14 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.1;net452</TargetFrameworks>
+    <TargetFrameworks>netstandard1.5;net452</TargetFrameworks>
```
jmarolf: I'll need to have a long hard think about whether netstandard1.5 is what we want.  Existing libraries may be targeting netstandard1.1 of even portable7 if they are old enough.  What is our back-compat cut-off?

sharwell: > I'll need to have a long hard think about whether netstandard1.5 is what we want.

When you test, you'll be testing against .NET Framework 4.5.2 or .NET *Core* 1.0 at the earliest. Either way, the new target wouldn't be a problem.

PR #155, opened 2018-08-25
Combine solutions

sharwell:


sharwell:
@jmarolf I did not see a way to cleanly implement #156 without this approach. Can you suggest an alternative?

PR #156, opened 2018-08-25
Use new test helpers in samples

sharwell:


samples/CSharp/Analyzers/Analyzers.Test/Analyzers.CSharp.Tests.csproj
```diff
@@ -1,11 +1,16 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>net46</TargetFramework>
```
jmarolf: This should be unnecessary now that we target .NET Standard 2.0

sharwell:
> After a lot of back and forth on other options I think this is the right thing to do for now

I'm going to try and get #203 built and published before merging this so we can keep the target framework as netcoreapp2.0.

samples/CSharp/Analyzers/Analyzers.Test/Analyzers.CSharp.Tests.csproj
sharwell: This change was reverted.

PR #158, opened 2018-08-25
Fix namespaces of VB test types

sharwell:


jmarolf:
Visual Basic places classes automatically within the default namespace specified in the project file.  The other changes that move classes to separate files seem fine though.

sharwell:
> Visual Basic places classes automatically within the default namespace specified in the project file.

This was not the case when I used the output assemblies in other projects.

jmarolf:
> This was not the case when I used the output assemblies in other projects.

yep, I just confirmed you are correct

sharwell:
I'm happy to revert this if we find out in the future it was a configuration mistake or there is some better way to do the same thing. :+1:

PR #161, opened 2018-08-25
Fix incorrect markup handling in code fix test scenarios

sharwell:


src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Codefix.Testing/CodeFixTest`1.cs
```diff
@@ -210,11 +214,8 @@ private bool HasFixableDiagnostics()
                 return false;
             }
 
-            Verify.True(FixedSources.Count == 0 || (FixedSources.Count == 1 && IsNullOrEmpty(FixedSources[0].content)), $"'{nameof(FixedSources)}' not equal to '{nameof(TestSources)}'.");
```
sharwell: 📝 Allowing an empty fixed file to be treated as a "file not changed by code fix" scenario seems more confusing than valuable. This provision was removed during the refactoring.

PR #169, opened 2018-08-27
Use the TVerifier instance instead of throwing Exception

sharwell:


src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
```diff
@@ -376,70 +376,34 @@ protected async Task VerifyFixAsync(CancellationToken cancellationToken)
                     message.Append("\r\n");
                 }
 
-                throw new Exception(message.ToString());
+                Verify.True(false, message.ToString());
```
jmarolf: Can we add a Verify.Fail(message.ToString()); method? MSTest and NUnit have Assert.Fail even if xUnit does not.

sharwell: ➡️ Fixed in ace2985. The final commit d8f9b46 just reorders the `True` method within the verifier types; I can remove this commit if you want.

PR #170, opened 2018-08-27
Remove the Visual Studio 2015 item and project templates

sharwell:


jmarolf:
@sharwell can you also remove all the now-unused properties from Versions.props

PR #182, opened 2018-08-30
Fix the default behavior of NumberOfIncrementalIterations

sharwell:
This commit updates the handling of an unspecified `NumberOfIncrementalIterations` adhere to the documented behavior. Rather than apply a fixed limit to the number of allow iterations, the code now sets the limit according to the number of fixable diagnostics appearing in the input source.

### DIFF
--- a/tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing.UnitTests/CodeFixIterationTests.cs
+++ b/tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing.UnitTests/CodeFixIterationTests.cs
@@ -1,0 +1,355 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Composition;
+using System.Threading;
+using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing.UnitTests/CodeFixIterationTests.cs
```diff
@@ -0,0 +1,355 @@
...
+                return Task.CompletedTask;
+            }
+
+            private async Task<Document> CreateChangedDocument(Document document, TextSpan sourceSpan, CancellationToken cancellationToken)
+            {
+                var tree = await document.GetSyntaxTreeAsync(cancellationToken);
+                var root = await tree.GetRootAsync(cancellationToken);
+                var token = root.FindToken(sourceSpan.Start);
+                var replacement = int.Parse(token.ValueText) + 1;
+                var newToken = SyntaxFactory.Literal(token.LeadingTrivia, " " + replacement.ToString(), replacement, token.TrailingTrivia);
```
sharwell: 📝 The addition of a space here is due to the way we detect if the application of a code fix changed the diagnostics. The current algorithm assumes that the same diagnostic (by ID) appearing in the same location means the code fix iteration process has reached a steady-state. Obviously this is not correct for the sample diagnostic/fix created for this test, and it's reasonable to believe other analyzers could behave in a similar manner. I'll file a follow-up issue to address this.

PR #197, opened 2018-09-04
Enable Public API analyzer

### LINKED ISSUE #183: Add Public API Analyzer for M.CA.Testing
We need to be notified if we change a public API here.

sharwell:
Fixes #183


RoslynSDK.ruleset
```diff
@@ -16,4 +16,7 @@
     <Rule Id="SX1101" Action="Warning" />
     <Rule Id="SX1309" Action="Warning" />
   </Rules>
+  <Rules AnalyzerId="Roslyn.Diagnostics.Analyzers" RuleNamespace="Roslyn.Diagnostics.Analyzers">
+    <Rule Id="RS0026" Action="None" /> <!-- AvoidMultipleOverloadsWithOptionalParameters -->
```
sharwell: ❓ @jmarolf should we file a bug to resolve this?

jmarolf: Yes

PR #221, opened 2018-09-11
Rename VerifyExclusions to Exclusions and make it an enum

sharwell:
This change is motivated by two items:

1. Feedback from @AArnott suggesting the name `VerifyExclusions` was not especially clear
2. Bug DotNetAnalyzers/StyleCopAnalyzers#1701, where a diagnostic cannot be suppressed

* [x] Rename `VerifyExclusions` and make it an enum that can be expanded
* [x] Add `AnalysisExclusions.Suppression`, and by default verify that diagnostics are suppressed if the appropriate directive appears at the beginning of the file
* [x] Add tests for `AnalysisExclusions.Suppression`


sharwell:
📝 I have to force-push to this and wait for the builds to complete due to #222

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalysisExclusions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        /// <summary>
+        /// Analysis will not report diagnostics in generated code.
+        /// </summary>
+        GeneratedCode = 0x01,
+
+        /// <summary>
+        /// Diagnostics will not be reported if a <c>#pragma warning disable</c> appears at the beginning of the file.
+        /// </summary>
+        Suppression = 0x02,
```
AArnott: When is this ever *not* the case? Do analyzers even have the choice to show diagnostics in spite of such a `#pragma`? Or is the question whether the analyzer notices the `#pragma` and saves itself the effort of producing diagnostics that it figures Roslyn will filter out anyway?

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -98,10 +98,10 @@ public string TestCode
         public Dictionary<string, string> XmlReferences { get; } = new Dictionary<string, string>();
 
         /// <summary>
-        /// Gets or sets a value indicating whether exclusions for generated code should be tested automatically. The
-        /// default value is <see langword="true"/>.
+        /// Gets or sets a value the analysis exclusions to verify. The default value is a combination of
+        /// <see cref="AnalysisExclusions.GeneratedCode"/> and <see cref="AnalysisExclusions.Suppression"/>.
         /// </summary>
-        public bool VerifyExclusions { get; set; } = true;
+        public AnalysisExclusions Exclusions { get; set; } = AnalysisExclusions.GeneratedCode | AnalysisExclusions.Suppression;
```
AArnott: I find `AnalysisExclusions.Suppression` a peculiar flag to include as a default, since I've never written or seen an analyzer that proactively looks for `#pragma` and turns itself off. But if Roslyn is doing the filtering for the analyzer, why is this one even an option, since an analyzer can't avoid it?

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -163,8 +163,19 @@ protected async Task VerifyDiagnosticsAsync((string filename, SourceText content
                         .Select(x => IsInSourceFile(x, sources) ? x.WithLineOffset(1) : x)
                         .ToArray();
 
-                    var commentPrefix = Language == LanguageNames.CSharp ? "//" : "'";
-                    VerifyDiagnosticResults(await GetSortedDiagnosticsAsync(sources.Select(x => (x.filename, x.content.Replace(new TextSpan(0, 0), $" {commentPrefix} <auto-generated>\r\n"))).ToArray(), additionalFiles, additionalMetadataReferences, analyzers, cancellationToken).ConfigureAwait(false), analyzers, expectedResults);
+                    if (Exclusions.HasFlag(AnalysisExclusions.GeneratedCode))
```
AArnott: If this flag *isn't* set, is it worthwhile to verify that the generated code in fact *does* produce the warnings?

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalysisExclusions.cs
sharwell: StyleCop Analyzers has a rule that currently fails this (bug linked in the initial PR comment). Providing this option allows StyleCop Analyzers to switch to the new testing framework and file the bug as a follow-up issue, rather than block adoption altogether.

sharwell: Also test inputs that contain a `#pragma warning enable` somewhere in the file might need to remove this flag, because it tests the case by adding the line to the top of the file.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
AArnott: I guess I still find this API confusing then. Does `Exclusions` means that the verifier will not test it, or that it *should* test that the analyzer *omits* diagnostics for the identified scenario?
Is there another word that doesn't leave the ambiguity? Something like `VerifyNoDiagnosticsFor` ?

PR #240, opened 2018-12-23
Give code fix verifier assertions diff output

### LINKED ISSUE #239: Analyzer test fx: code fix verification failures should show diff between actual vs. expected
When an assertion failure occurs because the expected code doesn't match the actual code after a code fix is applied, rather than show a big string with the expected code and a big string with the actual, it should show a single string with a "diff" between the two, so it's much easier to identify what went wrong.

For example, this is really hard for a human to identify what went wrong from the error message:

```
Failed   VSSDK006CheckServicesExistAnalyzerTests.FieldAssigned_GetService_ThenUsedElsewhereAsync
Error Message:
 content of 'Test0.cs' was expected to be '
using Microsoft;
using Microsoft.VisualStudio.Shell;
using Microsoft.VisualStudio.Shell.Interop;

class Test : Package {
    IVsBuildManagerAccessor svc;
    protected override void Initialize() {
        base.Initialize();
        this.svc = this.GetService(typeof(SVsBuildManagerAccessor)) as IVsBuildManagerAccessor;
        Assumes.Present(this.svc);
    }

    void Foo() {
        this.svc.BeginDesignTimeBuild();
    }
}
' but was '
using Microsoft;
using Microsoft.VisualStudio.Shell;
using Microsoft.VisualStudio.Shell.Interop;

class Test : Package {
    IVsBuildManagerAccessor svc;
    protected override void Initialize() {
        base.Initialize();
        this.svc = this.GetService(typeof(SVsBuildManagerAccessor)) as IVsBuildManagerAccessor;
        Assumes.Present(svc);
    }

    void Foo() {
        this.svc.BeginDesignTimeBuild();
    }
}
'
Expected: True
Actual:   False
```

Where is the difference between these almost identical code snippets?

Now compare with output like this:

```diff
 Failed   VSSDK006CheckServicesExistAnalyzerTests.FieldAssigned_GetService_ThenUsedElsewhereAsync
 Error Message:
  content of 'Test0.cs' did not match. Diff shown with expected as baseline:
 using Microsoft;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;

 class Test : Package {
     IVsBuildManagerAccessor svc;
     protected override void Initialize() {
         base.Initialize();
         this.svc = this.GetService(typeof(SVsBuildManagerAccessor)) as IVsBuildManagerAccessor;
-        Assumes.Present(this.svc);
+        Assumes.Present(svc);
     } 

     void Foo() {
         this.svc.BeginDesignTimeBuild();
     }
 }
```

AArnott:
This makes diagnosing the actual difference between actual and expected much easier.

Closes #239

### DIFF
--- a/src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/IVerifierExtensions.cs
+++ b/src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/IVerifierExtensions.cs
@@ -1,0 +1,59 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Text;
+using DiffPlex;
+using DiffPlex.DiffBuilder;
+using DiffPlex.DiffBuilder.Model;
+
+namespace Microsoft.CodeAnalysis.Testing
+{
+    /// <summary>
+    /// Extensions on the <see cref="IVerifier"/> interface.
+    /// </summary>
+    public static class IVerifierExtensions
+    {
+        /// <summary>
+        /// Asserts that two strings are equal, and prints a diff between the two if they are not.
+        /// </summary>
+        /// <param name="verifier">The verifier instance.</param>
+        /// <param name="expected">The expected string. This is presented as the "baseline/before" side in the diff.</param>
+        /// <param name="actual">The actual string. This is presented as the changed or "after" side in the diff.</param>
+        /// <param name="message">The message to precede the diff, if the values are not equal.</param>
+        public static void EqualOrDiff(this IVerifier verifier, string expected, string actual, string message = null)
+        {
+            Requires.NotNull(verifier, nameof(verifier));
+
+            if (expected != actual)
+            {
+                var diffBuilder = new InlineDiffBuilder(new Differ());
+                var diff = diffBuilder.BuildDiffModel(expected, actual);
+                var messageBuilder = new StringBuilder();
+                messageBuilder.AppendLine(
+                    string.IsNullOrEmpty(message)
+                        ? "Actual and expected values differ. Expected shown in baseline of diff:"
+                        : message);
+
+                foreach (var line in diff.Lines)
+                {
+                    switch (line.Type)
+                    {
+                        case ChangeType.Inserted:
+                            messageBuilder.Append("+");
+                            break;
+                        case ChangeType.Deleted:
+                            messageBuilder.Append("-");
+                            break;
+                        default:
+                            messageBuilder.Append(" ");
+                            break;
+                    }
+
+                    messageBuilder.AppendLine(line.Text);
+                }
+
+                verifier.Fail(messageBuilder.ToString());
+            }
+        }
+    }
+}
--- a/src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
+++ b/src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
@@ -19,5 +19,7 @@
     <PackageReference Include="Microsoft.VisualBasic" Version="10.0.1" />
     <PackageReference Include="Microsoft.VisualStudio.Composition" Version="15.6.36" />
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+    <!-- Use PrivateAssets=compile to avoid exposing our DiffPlex dependency downstream as public API. -->
+    <PackageReference Include="DiffPlex" Version="1.4.2" PrivateAssets="compile" />
   </ItemGroup>
 </Project>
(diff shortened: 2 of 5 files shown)


src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/IVerifierExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+                    messageBuilder.Append(message);
+                    messageBuilder.AppendLine();
+                }
+
+                foreach (var line in diff.Lines)
+                {
+                    switch (line.Type)
+                    {
+                        case ChangeType.Inserted:
+                            messageBuilder.Append("+ ");
```
sharwell: ❓ Why the extra space here? The diff format I'm used to is a single-character prefix followed by the content.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
```diff
@@ -19,5 +19,6 @@
     <PackageReference Include="Microsoft.VisualBasic" Version="10.0.1" />
     <PackageReference Include="Microsoft.VisualStudio.Composition" Version="15.6.36" />
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+    <PackageReference Include="DiffPlex" Version="1.4.2" PrivateAssets="compile" />
```
sharwell: ❓ Why the explicit `PrivateAssets` here?

tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing.UnitTests/AdditionalFilesFixTests.cs
```diff
@@ -77,7 +77,7 @@ public async Task TestMarkupDiagnosticFixedByAddingAdditionalFileFailsIfTextInco
                 }.RunAsync();
             });
 
-            Assert.Equal("content of 'File1.txt' was expected to be 'Wrong file text' but was 'File text'", exception.Message);
+            Assert.Equal($"content of 'File1.txt' did not match. Diff shown with expected as baseline:{Environment.NewLine}- Wrong file text{Environment.NewLine}+ File text{Environment.NewLine}", exception.Message);
```
sharwell: 💡 It seems like a verbatim string here would improve readability, but this could work too.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/IVerifierExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+        public static void EqualOrDiff(this IVerifier verifier, string expected, string actual, string message = null)
+        {
+            Requires.NotNull(verifier, nameof(verifier));
+
+            if (expected != actual)
+            {
+                var diffBuilder = new InlineDiffBuilder(new Differ());
+                var diff = diffBuilder.BuildDiffModel(expected, actual);
+                var messageBuilder = new StringBuilder();
+                if (!string.IsNullOrEmpty(message))
```
sharwell: 💡 Maybe have a default fallback message instead of just printing a diff without any context?

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
AArnott: Because DiffPlex is an implementation detail of the test framework -- it's not exposed in the public API, nor would I expect by referencing your nuget package that I'd suddenly get a DiffPlex namespace that I can use. You'd want to be able to replace the DiffPlex dependency with another one if you want without it "breaking" your downstream users.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/IVerifierExtensions.cs
AArnott: No great reason. It was just what I was already doing in my vs-mef testing framework. I'll remove the space.

tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing.UnitTests/AdditionalFilesFixTests.cs
AArnott: That solves the appveyor LF problem, but when you build and test on linux, it will still fail because `core.autocrlf=true` means LF on linux.

src/Tools/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
AArnott: Done

PR #268, opened 2019-03-04
Use codebase in pkgdef

jmarolf:
This should fix command registration

### DIFF
--- a/src/Templates/VS2017/Roslyn.SDK.VS2017.csproj
+++ b/src/Templates/VS2017/Roslyn.SDK.VS2017.csproj
@@ -11,7 +11,6 @@
     <!-- VS Insertion -->
     <VisualStudioInsertionComponent>Microsoft.CodeAnalysis.SDK</VisualStudioInsertionComponent>
     <Ngen>false</Ngen>
-    <UseCodeBase>true</UseCodeBase>
   </PropertyGroup>
   <!-- Setting Template Content to None -->
   <ItemGroup>
--- a/src/Templates/VS2017/Roslyn.SDK.VS2017.csproj
+++ b/src/Templates/VS2017/Roslyn.SDK.VS2017.csproj
@@ -11,6 +11,7 @@
     <!-- VS Insertion -->
     <VisualStudioInsertionComponent>Microsoft.CodeAnalysis.SDK</VisualStudioInsertionComponent>
     <Ngen>false</Ngen>
+    <UseCodeBase>true</UseCodeBase>
   </PropertyGroup>
   <!-- Setting Template Content to None -->
   <ItemGroup>
--- a/src/Tools/SyntaxVisualizer/Roslyn.SyntaxVisualizer.Extension/Roslyn.SyntaxVisualizer.Extension.csproj
+++ b/src/Tools/SyntaxVisualizer/Roslyn.SyntaxVisualizer.Extension/Roslyn.SyntaxVisualizer.Extension.csproj
@@ -15,6 +15,7 @@
     <ExtensionInstallationFolder>Microsoft\SyntaxVisualizer</ExtensionInstallationFolder>
     <VisualStudioInsertionComponent>Roslyn.SyntaxVisualizer.Extension</VisualStudioInsertionComponent>
     <Ngen>false</Ngen>
+    <UseCodeBase>true</UseCodeBase>
   </PropertyGroup>
 
   <!-- References -->


src/Templates/VS2017/Roslyn.SDK.VS2017.csproj
```diff
@@ -11,6 +11,7 @@
     <!-- VS Insertion -->
     <VisualStudioInsertionComponent>Microsoft.CodeAnalysis.SDK</VisualStudioInsertionComponent>
     <Ngen>false</Ngen>
+    <UseCodeBase>true</UseCodeBase>
```
sharwell: ❓ Why is this needed with `<GeneratePkgDefFile>false</GeneratePkgDefFile>`?

src/Tools/SyntaxVisualizer/Roslyn.SyntaxVisualizer.Extension/Roslyn.SyntaxVisualizer.Extension.csproj
```diff
@@ -15,6 +15,7 @@
     <ExtensionInstallationFolder>Microsoft\SyntaxVisualizer</ExtensionInstallationFolder>
     <VisualStudioInsertionComponent>Roslyn.SyntaxVisualizer.Extension</VisualStudioInsertionComponent>
     <Ngen>false</Ngen>
+    <UseCodeBase>true</UseCodeBase>
```
tmat: Unrelated to this change but ` <IsProductComponent>false</IsProductComponent>` seems wrong since the project is also setting `VisualStudioInsertionComponent`. Arcade will override `IsProductComponent` if `VisualStudioInsertionComponent` is set.

tmat: Another unrelated thing:

```
<!-- workaround for https://github.com/NuGet/Home/issues/5894 -->
  <Import Project="$(ArtifactsObjDir)$(AssemblyName)\*.props" Condition="'$(MSBuildProjectExtension)' == '.tmp_proj'" />	  <Import Project="$(ArtifactsObjDir)$(AssemblyName)\*.props" Condition="'$(MSBuildProjectExtension)' == '.tmp_proj'" />
```

```
 <!-- workaround for https://github.com/NuGet/Home/issues/5894 -->
  <Import Project="$(ArtifactsObjDir)$(AssemblyName)\*.targets" Condition="'$(MSBuildProjectExtension)' == '.tmp_proj'" />	  <Import Project="$(ArtifactsObjDir)$(AssemblyName)\*.targets" Condition="'$(MSBuildProjectExtension)' == '.tmp_proj'" />
```

shouldn't be needed and also has no effect on .NET Framework 4.7.2, since the WPF temp proj extension changed there. Arcade SDK implements a workaround for the WPF temp projects, so you should be able to just delete these.

tmat: I'm a bit confused here by `CreateVsixContainer` being `false`. This project doesn't build VSIX?

jmarolf: No, the Roslyn.SDK.2017 project embeds this dll inside of it.  It used to be a VSIX at some point.  I need to go refactor this to be more clear.

src/Templates/VS2017/Roslyn.SDK.VS2017.csproj
jmarolf: @tmat 	for the answer...

jmarolf: Well I'll take a stab at it.  We don't generate a pkgdef for this project but we do for its dependencies.  I'll do some testing, perhaps this is not required on this project.

src/Tools/SyntaxVisualizer/Roslyn.SyntaxVisualizer.Extension/Roslyn.SyntaxVisualizer.Extension.csproj
tmat: So, then should `VisualStudioInsertionComponent`, `Ngen`, `ExtensionInstallationFolder`, etc. be set?

jmarolf: Unsetting them causes the assemblies to be NGEN'd

tmat: Setting `VisualStudioInsertionComponent` does. If this project doesn't produce VSIX then it shouldn't set `VisualStudioInsertionComponent`.

PR #285, opened 2019-03-29
Update DiffPlex to consider whitespace as differences

sharwell:


jmarolf:
@sharwell 	can you give an example of why this change is needed?

sharwell:
@jmarolf If a code fix makes a change and the indentation is wrong, the test will fail but the diff output will not show which lines resulted in the failure.

sharwell:
I'll try to get a test added that shows this.

sharwell:
@jmarolf I rebased this to add a test showing the behavior before this change, and then applied the change and updated the test to show the new expected results.

PR #291, opened 2019-04-02
Fail the build on reference to Microsoft.CodeAnalysis.Workspaces.Desktop

sharwell:


sharwell:
We need this assembly for `MefV1HostServices`.

PR #298, opened 2019-04-07
Rename AnalysisExclusions to TestBehaviors

sharwell:
Inverting the flags allows future additions to this enumeration without invalidating code that sets the property using the object initializer syntax in tests.


jmarolf:
Could we call these `AnalysisOptions` instead of `Quirks`?

sharwell:
@jmarolf We could. The reason I somewhat like 'quirks' is these typically represent deviations from the most common (and expected) analyzer patterns.

jmarolf:
I guess I feel quirks is a little presumptive.  Yes this is a much more advanced option but for someone that needs to verify this sort of thing this is not a quirk. As you say `AnalysisOptions` is probably not good enough. Perhaps just `AdditionalOptions`?

sharwell:
After some brainstorming we settled on `TestBehaviors`

PR #299, opened 2019-04-07
Provide a default EmptyDiagnosticAnalyzer

sharwell:
The empty analyzer is useful for testing code fixes for compiler warnings and errors.


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/EmptyDiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.CodeAnalysis.Testing
+{
+    /// <summary>
+    /// Defines a <see cref="DiagnosticAnalyzer"/> which does not report any diagnostics.
+    /// </summary>
+#pragma warning disable RS1001 // Missing diagnostic analyzer attribute.
```
jmarolf: I would add this to a suppression file as we do not intend to fix (as it is not a bug) this so the pragma is just noise

sharwell: I'm not a fan of suppression files. Any objection to using 'SuppressMessage' here so I can provide the explanation inline?

PR #324, opened 2019-05-08
Markup options

sharwell:
* Add MarkupOptions, and allow option ~~PreferFirstDescriptor~~ UseFirstDescriptor
* Implement DiagnosticLocationOptions.IgnoreAdditionalLocations

IgnoreAdditionalLocations was implemented based on experiences with dotnet/roslyn-analyzers#2400. The introduction of an additional location meant markup could no longer be used for code fix testing.

~~PreferFirstDescriptor~~ UseFirstDescriptor was implemented as an aid to testing code style analyzers in Roslyn, which declare support for multiple diagnostics but nearly always just use the first descriptor.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/MarkupOptions.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/MarkupOptions.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.CodeAnalysis.Testing
+{
+    /// <summary>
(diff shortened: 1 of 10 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/MarkupOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// No additional markup options are specified.
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Use the first matching diagnostic descriptor when multiple diagnostics match the syntax. By default, this
+        /// option is not specified and the markup parser will fail when the syntax does not represent a <em>unique</em>
+        /// descriptor.
+        /// </summary>
+        PreferFirstDescriptor = 0x0001,
```
jmarolf: we aren't really preferring it perhaps `UseFirstDescriptor` makes more sense?

PR #328, opened 2019-05-28
Initial support for project references

sharwell:


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ProjectState.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Microsoft.CodeAnalysis.Testing
+{
+    public abstract class ProjectState
+    {
+        private readonly string _name;
+        private readonly string _defaultPrefix;
+        private readonly string _defaultExtension;
+
+        protected ProjectState(string name, string defaultPrefix, string defaultExtension)
```
jmarolf: nit: Perhaps call this `ProjectData``?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ProjectState.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            _name = name;
+            _defaultPrefix = defaultPrefix;
+            _defaultExtension = defaultExtension;
+
+            Sources = new SourceFileList(defaultPrefix, defaultExtension);
+        }
+
+        public string Name { get; set; }
```
jmarolf: Its not obvious to me why you can set this value after creating the object

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ProjectState.cs
jmarolf: nevermind this is to match `SolutionState`.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ProjectState.cs
sharwell: I'll make it read only

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ProjectState.cs
sharwell: Yes. I have a good feeling this API will change forms (e.g. to support diagnostics/markup/Fix All in Solution testing), but I'm hoping it's a decent start to unblock some testing for @mavasani.

PR #375, opened 2019-08-07
Added wizard and nuget references to CSharp Analyzer Item Template

peteraritchie:
Addresses #374

### DIFF
--- a/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ItemTemplates/CSharp/Analyzer/CSharpAnalyzer.vstemplate
+++ b/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ItemTemplates/CSharp/Analyzer/CSharpAnalyzer.vstemplate
@@ -11,4 +11,14 @@
   <TemplateContent>
     <ProjectItem SubType="Code" TargetFileName="$fileinputname$.cs" ReplaceParameters="true">Analyzer.cs</ProjectItem>
   </TemplateContent>
+  <WizardExtension>
+    <Assembly>NuGet.VisualStudio.Interop, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</Assembly>
+    <FullClassName>NuGet.VisualStudio.TemplateWizard</FullClassName>
+  </WizardExtension>
+  <WizardData>
+    <packages>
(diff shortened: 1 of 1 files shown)


sharwell:
After this change, what will happen if the template is used in a project that has a reference to the 2.0 release of Microsoft.CodeAnalysis?

peteraritchie:
> After this change, what will happen if the template is used in a project that has a reference to the 2.0 release of Microsoft.CodeAnalysis?

I was assuming that templates intelligently install versions of nuget packages.  But, it's not appearing to be a correct assumption.  Can you confirm?

sharwell:
It's hard to tell from the [documentation](https://docs.microsoft.com/en-us/nuget/visual-studio-extensibility/visual-studio-templates#adding-packages-to-a-template) how this would behave. I'm not sure I'd be comfortable having the template force package upgrades, but at the same time I never use the templates myself so I'm not sure it matters.

jmarolf:
@peteraritchie 	this looks good.  Could you also add these nuget dependencies for the other item templates?

src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ItemTemplates/CSharp/Analyzer/CSharpAnalyzer.vstemplate
```diff
@@ -11,4 +11,14 @@
   <TemplateContent>
     <ProjectItem SubType="Code" TargetFileName="$fileinputname$.cs" ReplaceParameters="true">Analyzer.cs</ProjectItem>
   </TemplateContent>
+  <WizardExtension>
+    <Assembly>NuGet.VisualStudio.Interop, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</Assembly>
+    <FullClassName>NuGet.VisualStudio.TemplateWizard</FullClassName>
+  </WizardExtension>
+  <WizardData>
+    <packages>
+      <package id="Microsoft.CodeAnalysis.Workspaces.Common" version="3.2.0" />
```
sharwell, line 20: ❗ The workspaces layer cannot be added to an analyzer project. Code fixes (which use the workspaces layer) go in a different assembly from analyzers.

```suggestion
```

jmarolf:
> I'm still uncomfortable with having the item templates changing package references.

Can you explain your objection? Other item templates do this

PR #377, opened 2019-09-06
Enable nullable reference types

sharwell:


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -588,7 +588,6 @@ private Solution GetSolution((string filename, SourceText content)[] sources, (s
         /// strings.</returns>
         protected Project CreateProject((string filename, SourceText content)[] sources, (string filename, SourceText content)[] additionalFiles, ProjectState[] additionalProjects, MetadataReference[] additionalMetadataReferences, string language)
         {
-            language = language ?? language;
```
jmarolf: Why the behavior change?

jmarolf: actually I have no idea what this was trying to do before

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/TestFileMarkupParser.cs
```diff
@@ -204,6 +204,10 @@ public static void GetPositionAndSpans(string input, out int? cursorPositionOpt,
         public static void GetPositionAndSpans(string input, out string output, out int cursorPosition, out IDictionary<string, IList<TextSpan>> spans)
         {
             GetPositionAndSpans(input, out output, out int? cursorPositionOpt, out spans);
+            if (cursorPositionOpt is null)
+            {
+                throw new InvalidOperationException();
```
jmarolf: Can we provide a better error message here to the user?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: > actually I have no idea what this was trying to do before

Exactly 😂

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/TestFileMarkupParser.cs
sharwell: Done!

Issue #237, opened 2018-11-28
CodeStyle tests fail on Mono


tmat:
```
System.TypeInitializationException : The type initializer for 'Microsoft.CodeAnalysis.Testing.MetadataReferences' threw an exception.\n---- System.IO.FileNotFoundException : Could not load file or assembly 'Microsoft.VisualBasic, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.
Stack trace
  at (wrapper managed-to-native) System.Object.__icall_wrapper_mono_generic_class_init(intptr)
  at Microsoft.CodeAnalysis.Testing.AnalyzerTest`1[TVerifier].CreateSolution (Microsoft.CodeAnalysis.ProjectId projectId, System.String language) [0x0008a] in <dd18822397554aca8b5e68ec5a8f6fe9>:0 
  at Microsoft.CodeAnalysis.Testing.AnalyzerTest`1[TVerifier].CreateProjectImpl (System.ValueTuple`2[System.String,Microsoft.CodeAnalysis.Text.SourceText][] sources, System.ValueTuple`2[System.String,Microsoft.CodeAnalysis.Text.SourceText][] additionalFiles, Microsoft.CodeAnalysis.MetadataReference[] additionalMetadataReferences, System.String language) [0x0001b] in <dd18822397554aca8b5e68ec5a8f6fe9>:0 
  at Microsoft.CodeAnalysis.Testing.AnalyzerTest`1[TVerifier].CreateProject (System.ValueTuple`2[System.String,Microsoft.CodeAnalysis.Text.SourceText][] sources, System.ValueTuple`2[System.String,Microsoft.CodeAnalysis.Text.SourceText][] additionalFiles, Microsoft.CodeAnalysis.MetadataReference[] additionalMetadataReferences, System.String language) [0x0000b] in <dd18822397554aca8b5e68ec5a8f6fe9>:0 
  at Microsoft.CodeAnalysis.Testing.AnalyzerTest`1[TVerifier].GetSolution (System.ValueTuple`2[System.String,Microsoft.CodeAnalysis.Text.SourceText][] sources, System.ValueTuple`2[System.String,Microsoft.CodeAnalysis.Text.SourceText][] additionalFiles, Microsoft.CodeAnalysis.MetadataReference[] additionalMetadataReferences) [0x00025] in <dd18822397554aca8b5e68ec5a8f6fe9>:0 
[...]

PR #388, opened 2019-09-16
Avoid trying to load Microsoft.VisualBasic when not necessary

sharwell:
This change allows analyzer tests for C# to run on Mono.

See #237
See mono/mono#10679

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/MetadataReferences.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/MetadataReferences.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using Microsoft.CodeAnalysis;
 
 #if NETSTANDARD1_5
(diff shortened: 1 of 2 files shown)


jmarolf:
@sharwell 	can we add a mono test leg to this repo to ensure we don't regress this?

sharwell:
@jmarolf We can add a Mono test leg, but even with that I'm not sure it would ensure no regressions. Mono does not support the API surface of net472 or netcoreapp2.2, so it is allowed to fail these tests at any time for any reason.

sharwell:
@jmarolf Note that I have no idea how to add a mono test leg, so we would need to get infrastructure and/or arcade support for that part.

jmarolf:
@sharwell 	I assume @tmat 	encountered this issue on an arcade repo so there must be a way?

jmarolf:
>Mono does not support the API surface of net472 or netcoreapp2.2, so it is allowed to fail these tests at any time for any reason.

I don't understand what we are doing then.  How is this not just an infinite bug tail for us?

bradwilson:
>I don't understand what we are doing then.

@sharwell is trying to make this usable for @xunit's analyzers, where Mono (on Linux/macOS) support is a hard requirement.

sharwell:
@bradwilson If this doesn't work, I'll take the extraction one step further to avoid having this reference in the code paths you're relying on.

PR #389, opened 2019-09-18
Use Visual Studio 2019 for builds

sharwell:


jmarolf:
I am fine with this change provided you can find an image with the 4.7.2 SDK installed on it.

PR #395, opened 2019-09-29
Add exception handler in DiagnosticResult.Message

NextTurn:
To be consistent with `DiagnosticResult.ToString`, `Diagnostic.GetMessage` and `LocalizableResourceString.GetText` when format arguments are not provided.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/DiagnosticResult.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/DiagnosticResult.cs
@@ -76,7 +76,14 @@ namespace Microsoft.CodeAnalysis.Testing
 
                 if (MessageFormat != null)
                 {
-                    return string.Format(MessageFormat.ToString(), MessageArguments ?? EmptyArguments);
+                    try
+                    {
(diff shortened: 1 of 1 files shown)


NextTurn:
Can you publish some newer packages to MyGet feed?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/DiagnosticResult.cs
```diff
@@ -76,7 +76,14 @@ public DiagnosticResult(DiagnosticDescriptor descriptor)
 
                 if (MessageFormat != null)
                 {
-                    return string.Format(MessageFormat.ToString(), MessageArguments ?? EmptyArguments);
```
sharwell: Since this no longer throws, do we still need the handler in GetMessage?

NextTurn:
@sharwell When will some new packages be published?

sharwell:
@NextTurn the new packages are automatically published to myget after merging 👍

NextTurn:
Ah, I see them. But the newer versions `1.0.0-beta1.*` are less than the older versions `1.0.0-beta1-*`, so the `1.0.0-beta1-63812-02` package published in Feb 2019 is considered as the latest one 😰. You may need to update the version format.

sharwell:
Totally agree /cc @JoeRobich

PR #401, opened 2019-11-04
Fix failure to include all reference assemblies

sharwell:


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -334,7 +367,6 @@ private async Task<ImmutableArray<MetadataReference>> ResolveCoreAsync(string la
                             if (File.Exists(Path.Combine(installedFrameworkPath, ReferenceAssemblyPath, item + ".dll")))
                             {
                                 resolvedAssemblies.Add(Path.GetFullPath(Path.Combine(installedFrameworkPath, ReferenceAssemblyPath, item + ".dll")));
-                                break;
```
sharwell: 📝 These `break` statements were added for a previous change when nested loops were used, but I forgot to remove them in a98e27c80362291cffa85e0fb095f33970005076 when I removed the inner loop.

PR #402, opened 2019-11-04
Update Versions.props

RikkiGibson:
Assuming this will fix the issue where our newer packages are not actually considered to be a higher version by NuGet. Follow up on https://github.com/dotnet/roslyn-sdk/pull/395#issuecomment-540862477

/cc @sharwell, @JoeRobich.


sharwell:
You should get @jmarolf to review this as well.

PR #403, opened 2019-11-04
Fix netstandard2.0

sharwell:
I found these issues while updating xunit.analyzers to use the new reference assemblies feature. Since xunit.assert doesn't multitarget netstandard2.0, it relies on facade assemblies from netstandard1.x to resolve types during the build. For this package, the facade assemblies are in the main references folder instead of a "Facade" subfolder.

In addition, we need to include the reference assembly package in the resolver context to make sure the resolver is aware of the correct library. In the xunit.analyzer case, the resolver was using NETStandard.Library 1.6.1 and all its dependencies even though NETStandard.Library 2.0.3 was the preferred version.


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -739,8 +755,121 @@ public static class NetStandard
                     new PackageIdentity(
                         "NETStandard.Library",
                         NuGetVersion.Parse("2.0.3")),
-                    @"build\netstandard2.0\ref")
-                .AddAssemblies(ImmutableArray.Create("netstandard"));
+                    Path.Combine("build", "netstandard2.0", "ref"))
+                .AddAssemblies(ImmutableArray.Create("netstandard"))
+                .AddAssemblies(ImmutableArray.Create(
+                    "Microsoft.Win32.Primitives",
```
mavasani: Is this the only reasonable way to handle this? Will this be scalable?

sharwell: It's hard-coded in the NuGet package using a custom targets file, rather than using the standard patterns.

PR #405, opened 2019-11-06
Fix package extraction in net452 builds

sharwell:
Previously, these builds were not extracting the .nuspec file for packages, which led to exceptions in the PackageFolderReader.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
@@ -243,7 +243,11 @@ namespace Microsoft.CodeAnalysis.Testing
                 Directory.CreateDirectory(temporaryPackagesFolder);
                 var localPathResolver = new PackagePathResolver(temporaryPackagesFolder);
 #if NET452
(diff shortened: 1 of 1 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -243,7 +243,11 @@ private async Task<ImmutableArray<MetadataReference>> ResolveCoreAsync(string la
                 Directory.CreateDirectory(temporaryPackagesFolder);
                 var localPathResolver = new PackagePathResolver(temporaryPackagesFolder);
 #if NET452
-                var packageExtractionContext = new PackageExtractionContext(logger);
+                var packageExtractionContext = new PackageExtractionContext(logger)
+                {
+                    PackageSaveMode = PackageSaveMode.Defaultv3,
+                    XmlDocFileSaveMode = XmlDocFileSaveMode.None,
+                };
```
jmarolf: Is there a way to add a test for this case?

sharwell: It would be difficult to create a test for this case, as it would depend on the cache being empty at the start of the test, and not populated by tests running for another TFM in the same run.

PR #406, opened 2019-11-06
Avoid directly exposing NuGet APIs

sharwell:
NuGet APIs tend to change in ways that do not support the high-level compatibility approach of Microsoft.CodeAnalysis.Testing. This change wraps the PackageIdentity class to avoid directly exposing the type to consuming code.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/PackageIdentity.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/PackageIdentity.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using NuGet.Versioning;
+
+namespace Microsoft.CodeAnalysis.Testing
+{
+    public sealed class PackageIdentity
+    {
+        public PackageIdentity(string id, string version)
+        {
+            Id = id ?? throw new ArgumentNullException(nameof(id));
+            Version = version ?? throw new ArgumentNullException(nameof(version));
+        }
+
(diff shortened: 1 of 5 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/PackageIdentity.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            Id = id ?? throw new ArgumentNullException(nameof(id));
+            Version = version ?? throw new ArgumentNullException(nameof(version));
+        }
+
+        public string Id { get; }
+
+        public string Version { get; }
+
+        internal NuGet.Packaging.Core.PackageIdentity ToNuGetIdentity()
```
jmarolf: Explain how this won't throw if a newer version of nuget comes along

sharwell: It would still throw if the versions are mismatched. However, by restricting the NuGet references to one assembly, there is a lower chance that downstream projects will introduce their own references to NuGet types that cannot be forwarded.

sharwell: StyleCop Analyzers runs a set of tests for C# 8 with Roslyn 3.3.1. This assembly references another test assembly originally built against Roslyn 1.2.1, and runs them in the new context. The C# 8 assembly targeted net472 and picked up the correct (expected) versions of the testing library and the NuGet assemblies, but failed to account for the fact that the C# 6 test assembly had references into an older version of the NuGet assemblies that could not be addressed with binding redirection. This change eliminates the need for the C# 6 test assembly to directly reference NuGet APIs, thus allowing the testing library and NuGet APIs to be versioned as a pair.

PR #421, opened 2020-01-16
Automatically include all compiler diagnostics with IDs that match a provided code fix

### LINKED ISSUE #419: Automatically include all compiler diagnostics with IDs that match a provided code fix
It's currently more difficult than expected to write a test for a code fix for a compiler suggestion or warning (#418). It would be good to automatically include all compiler diagnostics matching IDs supported by the code fix provider(s) used in the test.

taori:
this PR fixes #419

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
@@ -782,11 +782,16 @@ namespace Microsoft.CodeAnalysis.Testing
                 var compilationWithAnalyzers = compilation.WithAnalyzers(analyzers, GetAnalyzerOptions(project), cancellationToken);
                 var allDiagnostics = await compilationWithAnalyzers.GetAllDiagnosticsAsync().ConfigureAwait(false);
 
-                diagnostics.AddRange(allDiagnostics.Where(diagnostic => IsCompilerDiagnosticIncluded(diagnostic, compilerDiagnostics)));
+                diagnostics.AddRange(allDiagnostics.Where(diagnostic => !IsCompilerDiagnostic(diagnostic) || IsCompilerDiagnosticIncluded(diagnostic, compilerDiagnostics)));
             }
 
             var results = SortDistinctDiagnostics(diagnostics);
             return results.ToImmutableArray();
+        }
+
+        private static bool IsCompilerDiagnostic(Diagnostic diagnostic)
+        {
+            return diagnostic.Descriptor.CustomTags.Contains(WellKnownDiagnosticTags.Compiler);
         }
 
         /// <summary>
@@ -797,34 +802,23 @@ namespace Microsoft.CodeAnalysis.Testing
         /// <returns>return true to exclude a diagnostic, false to leave it up to internal logic.</returns>
         protected virtual bool IsCompilerDiagnosticIncluded(Diagnostic diagnostic, CompilerDiagnostics compilerDiagnostics)
         {
-            return !IsCompilerDiagnostic(diagnostic)
-                || IsCompilerDiagnosticIncluded(diagnostic, compilerDiagnostics);
-
-            static bool IsCompilerDiagnostic(Diagnostic diagnostic)
+            switch (compilerDiagnostics)
             {
-                return diagnostic.Descriptor.CustomTags.Contains(WellKnownDiagnosticTags.Compiler);
-            }
+                case CompilerDiagnostics.None:
+                default:
+                    return false;
 
-            static bool IsCompilerDiagnosticIncluded(Diagnostic diagnostic, CompilerDiagnostics compilerDiagnostics)
-            {
-                switch (compilerDiagnostics)
-                {
-                    case CompilerDiagnostics.None:
-                    default:
-                        return false;
+                case CompilerDiagnostics.Errors:
+                    return diagnostic.Severity >= DiagnosticSeverity.Error;
 
-                    case CompilerDiagnostics.Errors:
-                        return diagnostic.Severity >= DiagnosticSeverity.Error;
+                case CompilerDiagnostics.Warnings:
+                    return diagnostic.Severity >= DiagnosticSeverity.Warning;
 
-                    case CompilerDiagnostics.Warnings:
-                        return diagnostic.Severity >= DiagnosticSeverity.Warning;
+                case CompilerDiagnostics.Suggestions:
+                    return diagnostic.Severity >= DiagnosticSeverity.Info;
 
-                    case CompilerDiagnostics.Suggestions:
-                        return diagnostic.Severity >= DiagnosticSeverity.Info;
-
-                    case CompilerDiagnostics.All:
-                        return true;
-                }
+                case CompilerDiagnostics.All:
+                    return true;
             }
         }
 
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
@@ -781,7 +781,8 @@ namespace Microsoft.CodeAnalysis.Testing
                 var compilation = await project.GetCompilationAsync(cancellationToken).ConfigureAwait(false);
                 var compilationWithAnalyzers = compilation.WithAnalyzers(analyzers, GetAnalyzerOptions(project), cancellationToken);
                 var allDiagnostics = await compilationWithAnalyzers.GetAllDiagnosticsAsync().ConfigureAwait(false);
-                diagnostics.AddRange(allDiagnostics.Where(IsIncludedDiagnostic));
+
+                diagnostics.AddRange(allDiagnostics.Where(diagnostic => IsCompilerDiagnosticIncluded(diagnostic, CompilerDiagnostics)));
             }
 
             var results = SortDistinctDiagnostics(diagnostics);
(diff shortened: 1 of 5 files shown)


taori:
If you prefer a compositional approach, feel free to let me know. I tried to keep the API changes small.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -822,6 +823,13 @@ bool IsCompilerDiagnosticIncluded(Diagnostic diagnostic)
             }
         }
 
+        /// <summary>
+        /// Extension point to allow inheriting classes to include diagnostics which would otherwise be filtered. The default implementation returns false
+        /// </summary>
+        /// <param name="diagnostic">the diagnostic which is subject to filter processes.</param>
+        /// <returns>return true to exclude a diagnostic, false to leave it up to internal logic.</returns>
+        protected internal virtual bool IsDiagnosticHandled(Diagnostic diagnostic) => false;
```
sharwell: ❔ Can this be just `protected`?

sharwell: 💭 What if we extract the current `IsCompilerDiagnosticIncluded` method to be a virtual method here? Then the derived type could be:

```
protected override bool IsCompilerDiagnosticInclude(Diagnostic diagnostic, CompilerDiagnostics compilerDiagnostics)
{
  if (base.IsCompilerDiagnosticIncluded(diagnostic, compilerDiagnostic))
    return true;

  // return result of fixable diagnostic ID check
}
```

taori: My bad, makes sense. Otherwise it's a pretty bad extension point.

taori: Yeah, that was my first thought, but i figured i'd pick a way which can't break the functionality by overriding, so i kept the changes as small as possible. 

I also thought of adding `List<IDiagnosticFilter> DiagnosticFilters` to `AnalyzerTest` so it is composable. I'll change it as per your request though, because anyone who works with this library, probably knows well enough why his tests suddenly stop working after changing the implementation.

taori:
@sharwell code is update as requested - one question though: How are reviews usually handled? Should i resolve them after updating the code as per review, or is that something which is usually done by the person opening a review?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -781,43 +781,49 @@ protected async Task<ImmutableArray<Diagnostic>> GetSortedDiagnosticsAsync(Solut
...
-            // Local function
-            bool IsIncludedDiagnostic(Diagnostic diagnostic)
-            {
-                return !IsCompilerDiagnostic(diagnostic)
-                    || IsCompilerDiagnosticIncluded(diagnostic);
-            }
+        /// <summary>
+        /// Extension point to allow inheriting classes to include diagnostics which would otherwise be filtered. The default implementation returns false
+        /// </summary>
+        /// <param name="diagnostic">the diagnostic which is subject to filter processes.</param>
```
sharwell:
```suggestion
        /// <param name="diagnostic">The compiler diagnostic.</param>
```

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -781,43 +781,49 @@ protected async Task<ImmutableArray<Diagnostic>> GetSortedDiagnosticsAsync(Solut
...
+        /// <param name="compilerDiagnostics">level of diagnostic used to filter compiler diagnostics</param>
```
sharwell:
```suggestion
        /// <param name="compilerDiagnostics">The compiler diagnostic level in effect for the test.</param>
```

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -781,43 +781,49 @@ protected async Task<ImmutableArray<Diagnostic>> GetSortedDiagnosticsAsync(Solut
...
+        /// <returns>return true to exclude a diagnostic, false to leave it up to internal logic.</returns>
```
sharwell:
```suggestion
        /// <returns><see langword="true"/> to include the diagnostic for validation; otherwise, <see langword="false"/> to exclude a diagnostic.</returns>
```

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -781,43 +781,49 @@ protected async Task<ImmutableArray<Diagnostic>> GetSortedDiagnosticsAsync(Solut
...
+        protected virtual bool IsCompilerDiagnosticIncluded(Diagnostic diagnostic, CompilerDiagnostics compilerDiagnostics)
+        {
+            return !IsCompilerDiagnostic(diagnostic)
```
sharwell: 📝 The `!IsCompilerDiagnostic(diagnostic)` should still be up in the caller. When `IsCompilerDiagnosticIncluded` is called, we expect the argument to always (already) be a compiler diagnostic.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -781,43 +781,49 @@ protected async Task<ImmutableArray<Diagnostic>> GetSortedDiagnosticsAsync(Solut
...
+        }
 
-            // Local function
-            bool IsIncludedDiagnostic(Diagnostic diagnostic)
-            {
-                return !IsCompilerDiagnostic(diagnostic)
-                    || IsCompilerDiagnosticIncluded(diagnostic);
-            }
+        /// <summary>
+        /// Extension point to allow inheriting classes to include diagnostics which would otherwise be filtered. The default implementation returns false
```
sharwell:
```suggestion
        /// Determines if a compiler diagnostic should be included for diagnostic validation. The default implementation includes all diagnostics at a severity level indicated by <paramref name="compilerDiagnostics"/>.
```

sharwell:
> Should i resolve them after updating the code as per review, or is that something which is usually done by the person opening a review?

It tends to vary. You can go with whatever approach you're comfortable with.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
taori: Mhm, i suppose i changed that after you started the review. At least without exclamation mark tests failed and i restored the code to the change you mention here.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -781,43 +781,49 @@ protected async Task<ImmutableArray<Diagnostic>> GetSortedDiagnosticsAsync(Solut
                 var compilation = await project.GetCompilationAsync(cancellationToken).ConfigureAwait(false);
                 var compilationWithAnalyzers = compilation.WithAnalyzers(analyzers, GetAnalyzerOptions(project), cancellationToken);
                 var allDiagnostics = await compilationWithAnalyzers.GetAllDiagnosticsAsync().ConfigureAwait(false);
-                diagnostics.AddRange(allDiagnostics.Where(IsIncludedDiagnostic));
+
+                diagnostics.AddRange(allDiagnostics.Where(diagnostic => IsCompilerDiagnosticIncluded(diagnostic, compilerDiagnostics)));
```
sharwell: Regarding the previous comment, here's what I expected:

```suggestion
                diagnostics.AddRange(allDiagnostics.Where(diagnostic => !IsCompilerDiagnostic(diagnostic) || IsCompilerDiagnosticIncluded(diagnostic, compilerDiagnostics)));
```

taori: Ohhh okay. I misunderstood. I'll do it manually then so the method is in scope to be callable.

taori:
@sharwell if this were to be merged now that the PR is ready again, would it end up straight in the beta channel?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: This is still not resolved

sharwell:
@taori Yes, it will automatically go to MyGet after it's merged.

sharwell:
I'm just waiting on the refactoring so `IsCompilerDiagnostic` is called before `IsCompilerDiagnosticIncluded` is called.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
taori: thanks for the reminder, forgot to push my last commit

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
```diff
@@ -169,6 +169,24 @@ protected CodeFixTest()
...
 
+        /// <inheritdoc />
+        protected override bool IsCompilerDiagnosticIncluded(Diagnostic diagnostic, CompilerDiagnostics compilerDiagnostics)
+        {
+            if (base.IsCompilerDiagnosticIncluded(diagnostic, compilerDiagnostics))
+            {
+                return true;
+            }
+
+            return CodeFixProvidersHandleDiagnostic(diagnostic);
```
jmarolf: Why does this need to be a local function?

sharwell: It doesn't need to be, but it's not hurting anything either. I merged so @taori can start using this 😄

taori: Original function combined multiple local functions. Should have refactored that one out too, like other methods.

sharwell:
Thanks @taori !!

taori:
@sharwell Thanks for your guidance and feedback :)

PR #431, opened 2020-01-30
Exclude obj folders from git

### LINKED ISSUE #430: src/VisualStudio.Roslyn.SDK/Roslyn.SDK/obj/ is not excluded
Clone the repo, call `restore.cmd`, `build.cmd`. Now you have one folder marked as `untracked`.

Evangelink:
Fix #430

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 artifacts/
 .dotnet/
 .tools/
+obj/
 
 # ignore templates modified by build output
 src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ModifiedTemplates


.gitignore
```diff
@@ -18,6 +18,7 @@
 artifacts/
 .dotnet/
 .tools/
+obj/
```
Evangelink: Is it worth adding all the usual folders here?

sharwell:
There is no need to exclude this folder. This repository places all intermediate output under **artifacts**, which is already excluded. I'll put more information in the linked issue.

PR #436, opened 2020-02-04
Visual Studio Project Template: "Analyzer with Code Fix": Verifier does not properly account for multiple source files

JakenVeina:
Fixed that DiagnosticVerifier.Helper sorts result Diagnostic objects only by SourceSpan, meaning that when multiple source files are involved, VerifyDiagnostic and VerifyCodeFix can report false negatives, since they both rely on Diagnostic and DiagnosticResult objects being sorted the same, in order to successfully check for differences.


jmarolf:
@JakenVeina  Thanks so much!

sharwell:
@JakenVeina This is the old test framework. I would highly recommend switching to [Microsoft.CodeAnalysis.Testing](https://github.com/dotnet/roslyn-sdk/tree/master/src/Microsoft.CodeAnalysis.Testing) instead of using the old templates.

PR #451, opened 2020-02-14
Include a trailing comma in lists

sharwell:


jmarolf:
I don't understand the motivation for this one. What user experience is made better by this change?

sharwell:
I always have to add this comma when I copy/paste the list into my test initializer.

jmarolf:
Thats a good enough reason for me

PR #454, opened 2020-02-14
Handle reordering of expected arguments

sharwell:


tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.UnitTests/CompilerErrorTests.cs
```diff
@@ -72,6 +72,29 @@ class TestClass {
             }.RunAsync();
         }
 
+        [Fact]
+        public async Task TestCSharpReorderedExplicitCompilerErrorWithExplicitInterfaceSymbol()
```
sharwell: 📝 It took a while but I made sure to create a test that fails without this change

PR #455, opened 2020-02-16
Improved error reporting

sharwell:
* [x] Provide more complete information about context when comparisons fail
* [x] Fix failure to include additional locations in sample outputs
* [x] Implement partial match support for diagnostics, so the most common small errors are easier to resolve

The new backtracking algorithm for matching actual and expected diagnostics does a great job of finding matching combinations, but does a poor job of finding the right elements to compare when some items don't match. This change improves both match selection and error reporting for cases where diagnostics do not match.


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -435,7 +466,8 @@ int RecursiveMatch(int firstActualIndex, int firstExpectedIndex, int unmatchedAc
                     }
 
                     var (lineSpan, additionalLineSpans) = actualResultLocations[firstActualIndex];
-                    if (!IsMatch(actualResults[firstActualIndex], actualIds[firstActualIndex], lineSpan, additionalLineSpans, actualArguments[firstActualIndex], expectedResults[i], expectedArguments[i]))
+                    var matchValue = GetMatchValue(actualResults[firstActualIndex], actualIds[firstActualIndex], lineSpan, additionalLineSpans, actualArguments[firstActualIndex], expectedResults[i], expectedArguments[i]);
+                    if (matchValue == 1)
```
jmarolf: convice me that this comparison of int to double will work 100% of the time (or ad a comment about why its ok)

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -289,48 +289,75 @@ private void VerifyDiagnosticResults(IEnumerable<Diagnostic> actualResults, Immu
 
                 if (!expected.HasLocation)
                 {
-                    verifier.Equal(Location.None, actual.Location, $"Expected:\r\nA project diagnostic with No location\r\nActual:\r\n{FormatDiagnostics(analyzers, actual)}");
+                    message = $"Expected a project diagnostic with no location:{Environment.NewLine}" +
```
jmarolf: perhaps we can add some extensions to the `verifier` so we don't need to repeat all this message formatting code everywhere

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -621,7 +690,7 @@ private static string FormatDiagnostics(ImmutableArray<DiagnosticAnalyzer> analy
                 if (applicableAnalyzer != null)
                 {
                     var analyzerType = applicableAnalyzer.GetType();
-                    var rule = applicableAnalyzer.SupportedDiagnostics.Length == 1 ? string.Empty : $"{analyzerType.Name}.{diagnosticsId}";
+                    var rule = location != Location.None && location.IsInSource && applicableAnalyzer.SupportedDiagnostics.Length == 1 ? string.Empty : $"{analyzerType.Name}.{diagnosticsId}";
```
jmarolf: seems weird that we are getting `default` or not-in-source locations here. Why is this? Is there a better place to filter them out?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
jmarolf: As I recall mantissas etc, I remember that `0` and `1` are always safe to compare to int. So this is fine, but perhaps add a comment?

---
In reply to: [379929020](https://github.com/dotnet/roslyn-sdk/pull/455#discussion_r379929020) [](ancestors = 379929020)

sharwell: I introduced a new `MatchQuality` structure to avoid relying on `int`/`double`. I wouldn't say it's a perfect type now, but it should be much better for future refactoring.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: Extracted a helper method for this

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: The testing framework supports validation of no-location diagnostics. This line is just a minor change to failure messaging; if there is more to do here it seems reasonable to leave it as a follow-up.

PR #473, opened 2020-02-24
Add AnalyzerTest<TVerifier>.CreateParseOptions()

sharwell:


tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.UnitTests/AutoExclusionTests.cs
```diff
@@ -303,6 +303,11 @@ protected override CompilationOptions CreateCompilationOptions()
                 return new CSharp.CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary);
             }
 
+            protected override ParseOptions CreateParseOptions()
```
sharwell: 📝 Tests generally will not need to define this because it would be inherited from `CSharpAnalyzerTest`, `CSharpCodeFixTest`, or one of the VB equivalents. Most of these tests are in a UnitTests project that only references the language-independent test framework, so these helpers are not available.

PR #474, opened 2020-02-24
Extract base class CodeActionTest<TVerifier>

sharwell:
This base class simplifies an upcoming change to add `CodeRefactoringVerifier<>`.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest`1.cs
@@ -1,0 +1,391 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CodeActions;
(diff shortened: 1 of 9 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest`1.cs
```diff
@@ -0,0 +1,391 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the index of the code action to apply.
+        /// </summary>
+        /// <remarks>
+        /// <para>If <see cref="CodeActionIndex"/> and <see cref="CodeActionEquivalenceKey"/> are both specified, the
+        /// test will further verify that the two properties refer to the same code action.</para>
+        /// </remarks>
+        /// <seealso cref="CodeActionEquivalenceKey"/>
+        public int? CodeActionIndex { get; set; }
```
sharwell: 📝 The most significant change in this pull request is the renaming of `CodeFixIndex` to `CodeActionIndex`. This property is definitely in use; the change is not especially complicated but I can provide an `[Obsolete]` property if we want.

jmarolf: This library is _supposed_ to be pre-release :) so this should be fine

---
In reply to: [383459263](https://github.com/dotnet/roslyn-sdk/pull/474#discussion_r383459263) [](ancestors = 383459263)

sharwell: I added pass-through properties to aid the migration for a period of time.

PR #478, opened 2020-02-25
Improve reference assembly resolution time

sharwell:
Review commit-by-commit is recommended.


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -195,7 +201,12 @@ private async Task<ImmutableArray<MetadataReference>> ResolveCoreAsync(string la
 
             using (var cacheContext = new SourceCacheContext())
             {
+                var temporaryPackagesFolder = Path.Combine(Path.GetTempPath(), "test-packages");
```
jmarolf: should we give the user the option to specify this location? also, when are these resources cleaned up?

sharwell: ➡️ Filed #487

PR #480, opened 2020-02-25
Extract common code to local function

sharwell:


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -436,15 +429,21 @@ private async Task<ImmutableArray<MetadataReference>> ResolveCoreAsync(string la
...
-                static string? GetInstalledPath(PackagePathResolver resolver, NuGet.Packaging.Core.PackageIdentity id)
+                static string? GetInstalledPath(PackagePathResolver localPathResolver, PackagePathResolver globalPathResolver, NuGet.Packaging.Core.PackageIdentity packageIdentity)
                 {
-                    try
-                    {
-                        return resolver.GetInstalledPath(id);
-                    }
-                    catch (PathTooLongException)
+                    return GetInstalledPath(localPathResolver, packageIdentity)
+                        ?? GetInstalledPath(globalPathResolver, packageIdentity);
```
sharwell: 📝 This will be getting more complex for #478, so I'm extracting it to one location to simplify that review.

Issue #482, opened 2020-02-26
How are "any" framework references resolved?


sharwell:
Consider the following project:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netstandard1.3</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="2.8.2" />
  </ItemGroup>
</Project>
```

In this example, **Microsoft.CodeAnalysis.Common** brings in a reference to **Microsoft.CodeAnalysis.Analyzers** version 1.1.0. The nuspec for this package has a framework reference for System without specifying the target framework. According to the NuGet APIs I’ve been working with, this is an `Any,Version=v0.0` group which the test code is applying during package resolution for the TFM `netstandard1.3`. However, if I look at the **project.assets.json** file produced by NuGet for this project, I see that the System dependency has been omitted for the `netstandard1.3` TFM. This leaves me with two questions:

1. Why is NuGet package restore for the above project not including a `System` dependency for **Microsoft.CodeAnalysis.Analyzers**?
1. For the answer to (1), where is the filtering implemented in the NuGet code?

For reference, the nuspec for the problematic package looks like this:

```xml
<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
  <metadata>
    <id>Microsoft.CodeAnalysis.Analyzers</id>
    <version>1.1.0</version>

    <!-- ...other various properties... -->

    <frameworkAssemblies>
      <frameworkAssembly assemblyName="System" targetFramework="" />
    </frameworkAssemblies>
  </metadata>
</package>
```

PR #483, opened 2020-02-26
Ignore 'any' framework references

sharwell:
See #482

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
@@ -353,7 +353,9 @@ namespace Microsoft.CodeAnalysis.Testing
                         }
                     }
 
-                    if (nearestFramework is object)
+                    // Ignore framework references for 'Any,Version=v0.0' until we know how NuGet is supposed to handle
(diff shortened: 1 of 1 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -353,7 +353,9 @@ private async Task<ImmutableArray<MetadataReference>> ResolveCoreAsync(string la
                         }
                     }
 
-                    if (nearestFramework is object)
+                    // Ignore framework references for 'Any,Version=v0.0' until we know how NuGet is supposed to handle
+                    // them. https://github.com/dotnet/roslyn-sdk/issues/482
+                    if (nearestFramework is { IsAny: false })
```
sharwell: 📝 The alternative to this is to only ignore the references if there is no reference assembly package specified.

jmarolf: Can you file a bug on the nuget team and link it here?

sharwell: I filed #482 here and sent an email to the NuGet team to ask how we are supposed to resolve it.

PR #154, opened 2018-08-24
Update project templates to use the new test library

sharwell:
* [x] ~~Update the VS2015 templates~~ (VS2015 templates removed in #170) 
* [x] Update the VS2017 templates

PR #492, opened 2020-03-26
Fixing C# template

Pvlerick:
This fixes one of the issues in #493


src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Test/DiagnosticVerifier.Helper.cs
```diff
@@ -15,10 +15,10 @@ namespace TestHelper
     /// </summary>
     public abstract partial class DiagnosticVerifier
```
sharwell: 📝 This is a legacy test class that is no longer recommended for use. The new test library, which we would like to incorporate into the templates, does not include this file at all. See #154 for my initial work on this that stalled a while back.

jmarolf:
I am ok accepting this PR with the knowledge that we (hopefully) replace it with the new library in development very soon.

src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Test/DiagnosticVerifier.Helper.cs
Pvlerick: Ok, I see. However this is still what you get when you use the template from VS :-(

PR #497, opened 2020-04-03
Nested code actions

### LINKED ISSUE #460: Nested code actions are not supported

sharwell:
Fixes #460 

During the implementation of this pull request, I also found and filed https://github.com/dotnet/roslyn/issues/43044 and https://github.com/dotnet/roslyn-analyzers/issues/3475.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest`1.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using System.Reflection;
 using System.Text;
@@ -63,11 +64,11 @@ namespace Microsoft.CodeAnalysis.Testing
                 || !oldState.AdditionalFiles.SequenceEqual(newState.AdditionalFiles, SourceFileEqualityComparer.Instance);
         }
 
-        protected static CodeAction? TryGetCodeActionToApply(List<CodeAction> actions, int? codeActionIndex, string? codeActionEquivalenceKey, IVerifier verifier)
+        protected static CodeAction? TryGetCodeActionToApply(ImmutableArray<CodeAction> actions, int? codeActionIndex, string? codeActionEquivalenceKey, IVerifier verifier)
         {
             if (codeActionIndex.HasValue && codeActionEquivalenceKey != null)
             {
-                if (actions.Count <= codeActionIndex)
+                if (actions.Length <= codeActionIndex)
                 {
                     return null;
                 }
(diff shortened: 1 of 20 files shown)


tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing.UnitTests/MultipleCodeFixesTests.cs
```diff
@@ -75,6 +75,43 @@ class TestClass {{
...
+";
+            var fixedCode = $@"
+class TestClass {{
+  int field = 1;
+}}
+";
+
+            // The batch code fix provider does not support nested code actions.
+            // https://github.com/dotnet/roslyn/issues/43044
+            var batchFixedCode = testCode;
```
sharwell, line 95: 📝 This test could have been written two different ways to work around the bug:

1. Disable the Fix All test
2. Use a different expected outcome for the normal and Fix All tests (specifically, the result is the same as the input which means no change was applied)

I opted for the second approach for one important reason: this approach ensures that when we finally do update the tests to a version of Roslyn that supports the batch fixer in this scenario, the test will start to fail. This test will be correct at the earliest point where it is valid to do so (no earlier or later), even if the person who fixes the bug isn't the same person who eventually updates the dependency in this project.

/cc @Evangelink @bartonjs @stephentoub who may be interested in how I approached a test scenario like this.

tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing.UnitTests/MultipleCodeFixesTests.cs
```diff
@@ -75,6 +75,43 @@ class TestClass {{
...
+                ImmutableArray.Create(2),
+                ImmutableArray.Create(3));
+            await new CSharpTest(codeFixes, nested: true)
+            {
+                TestCode = testCode,
+                FixedCode = fixedCode,
+                BatchFixedState =
+                {
+                    Sources = { batchFixedCode },
+                    MarkupHandling = MarkupMode.Allow,
```
sharwell, line 109: 📝 This line shouldn't be necessary. I filed #498 with details on a fix.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest`1.cs
```diff
@@ -93,6 +94,37 @@ protected static bool HasAnyChange(SolutionState oldState, SolutionState newStat
             }
         }
 
+        protected virtual ImmutableArray<CodeAction> FilterCodeActions(ImmutableArray<CodeAction> actions)
+        {
+            var builder = actions.ToBuilder();
+            while (true)
```
jmarolf, line 100: Will this ever return if there are no nested code actions?

sharwell: Yes, it will return. `changesMade` is initialized to false, and the loop only continues when processing an action sets it to true.

PR #500, opened 2020-04-05
Ensure MatchDiagnostics completes in a reasonable time

sharwell:
Prior to fully optimizing this algorithm, this change ensures that tests
complete in a reasonable time even if an exact match for diagnostics
cannot be found.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
@@ -148,6 +148,12 @@ namespace Microsoft.CodeAnalysis.Testing
         public List<Func<Solution, ProjectId, Solution>> SolutionTransforms { get; } = new List<Func<Solution, ProjectId, Solution>>();
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -147,6 +147,12 @@ public string TestCode
         /// </summary>
         public List<Func<Solution, ProjectId, Solution>> SolutionTransforms { get; } = new List<Func<Solution, ProjectId, Solution>>();
 
+        /// <summary>
+        /// Gets or sets the timeout to use when matching expected and actual diagnostics. The default value is 2
+        /// seconds.
+        /// </summary>
+        protected TimeSpan MatchDiagnosticsTimeout { get; set; } = TimeSpan.FromSeconds(2);
```
jmarolf, line 154: How was this decided? 2 seconds seems _very_ long to me.

sharwell: This has only been hit a couple times since the feature was implemented, and in both cases it was during the development of a new test. The timeout is mostly about providing the most usable error message in the event of a test failure, so they are naturally short-lived scenarios.

PR #501, opened 2020-04-05
Add test projects for MSTest

sharwell:


sharwell:
This will take a bit of work, but should be ready soon. I need to find a copy of the arcade implementation for MSTest. I already found one for NUnit as a follow-up.

PR #502, opened 2020-04-07
Fix support for markup around CDATA elements

sharwell:


tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.UnitTests/MarkupParserTests.cs
```diff
@@ -277,5 +277,23 @@ public void OverlappingSpans4C()
             var y = Assert.Single(ys);
             Assert.Equal(TextSpan.FromBounds(6, 12), y);
         }
+
+        [Fact]
+        public void CDataMarkup()
+        {
+            var markup = "{|X:[|<![CDATA[|]|}text[|]]>|]";
```
mavasani, line 284: I had to read this few times to grok what it means :-)

sharwell: So did I (actually I didn't have the output text handy so I had to debug it to see). The expected output string isn't required for the test but helps the reader figure out the crazy.

PR #511, opened 2020-04-15
Source generator samples

chsienki:
Adds a sample generator project which contains three example generators:
- HelloWorldGenerator.cs : a very simple example of injecting code into a users compilation
- XmlSettingsGenerator.cs: an example of using additional files to create classes
- AutoNotifyGenerator.cs: a more involved sample that implements automatic `INotifyPropertyChanged` support, and shows how to use the `SyntaxReceiver` functionality. 

These are still a work in progress, and we expect them to change as we evolve the feature. In the long run these are expected to represent both simple how-to's and best practices with generators (we'll probably update the SG cookbook in the roslyn repo to start pointing here for example).

The project doesn't have tests yet. I suspect we'll want to build out some tooling like we provide for analyzers for testing generators.


jmarolf:
@chsienki I suggest we keep this in its own branch until the feature is complete. Then we merge this into master once its done. We can also work on the testing portion of this in that branch

chsienki:
@jmarolf Happy either way, I was originally going to leave it in a branch, but @jaredpar preferred it be in master, so we can easily point people to them (in @cartermp's blog post for instance)

jmarolf:
@chsienki [this link](https://github.com/dotnet/roslyn-sdk/tree/source-generator-samples/samples/CSharp/SourceGenerators/SourceGeneratorSamples) appears to work today. Would like to understand why that is not sufficient for @cartermp

jaredpar:
I would prefer they be in `master`. This makes them more discoverable and hopefully in turns encourages customers to interact with them. 

This is a samples section and these are samples of our work in progress. The compiler changes are in `master` even though they're not complete. Feels odd if the supporting samples weren't. If we want to make it clearer that they're not RTM we can make a nice README to discuss that.

jmarolf:
I would be fine with a readme explaining that these are a work-in-progress. I am just concerned that we would have a bunch of shipped samples next to some in-progress ones with no differentiation

chsienki:
Added a readme to explain they're in progress. Happy to play with the wording to get something we all like.

cartermp:
How do I build these samples? `install.ps1` just errors out:

image

samples/CSharp/SourceGenerators/SourceGeneratorSamples/SourceGeneratorSamples.csproj
```diff
@@ -0,0 +1,46 @@
...
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <RestoreAdditionalProjectSources>https://dotnet.myget.org/F/roslyn/api/v3/index.json ;$(RestoreAdditionalProjectSources)</RestoreAdditionalProjectSources>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!--<PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="2.9.8" />-->
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.6.0-3.20207.2" PrivateAssets="all" />
```
cartermp, line 33: This currently emits a NuGet warning about depending on Microsoft.CodeAnalysis.Analyzers, picking 3.0.0-beta2.final.

chsienki: Fixed.

chsienki:
The install.ps1 are for installing the analyzers as a package, but we don't support it for generators yet. Removed to clear up confusion.

samples/CSharp/SourceGenerators/SourceGeneratorSamples/HelloWorldGenerator.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    [Generator]
+    public class HelloWorldGenerator : ISourceGenerator
+    {
+        public void Execute(SourceGeneratorContext context)
+        {
+            // begin creating the source we'll inject into the users compilation
+            StringBuilder sourceBuilder = new StringBuilder(@"
+using System;
+namespace HelloWorldGenerator
```
cartermp, line 17: Might be better to have this be `HelloWorldGenerated` to distinguish from the name of the generator

cartermp:
In the README there should be instructions on how to install the sample generators into your own project. Since there's no example here showing consumption, folks will definitely want the prescriptive guidance.

chsienki:
@cartermp I Added a demo project that consumes the samples, and allows a 'one click' run experience to see what they do, and how to consume them.

I updated the readme to include how to build and consume in your own project.

samples/CSharp/SourceGenerators/SourceGeneratorSamples/HelloWorldGenerator.cs
chsienki: Updated.

samples/CSharp/SourceGenerators/README.md
```diff
@@ -0,0 +1,35 @@
...
+For more infomation on the Source Generators feature, see the [design document](https://github.com/dotnet/roslyn/blob/master/docs/features/source-generators.md).
+
+Prerequisites
+-----
+
+These samples require Visual Studio 16.6 or higher.
+
+Building the samples
+-----
+Open `SourceGenerators.sln` in Visual Studio or run `dotnet build` from the `\SourceGenerators` directory.
```
cartermp, line 15: I don't see `SourceGenerators.sln` in this PR or repo. Did you mean to open `Samples.sln`?

Samples.sln
```diff
@@ -111,6 +111,10 @@ Project("{778DAE3C-4631-46EA-AA77-85C1314464D9}") = "VisualBasicToCSharpConverte
 EndProject
 Project("{778DAE3C-4631-46EA-AA77-85C1314464D9}") = "VisualBasicToCSharpConverter.UnitTests", "samples\VisualBasic\VisualBasicToCSharpConverter\VisualBasicToCSharpConverter.Test\VisualBasicToCSharpConverter.UnitTests.vbproj", "{5B7D7569-B5EE-4C01-9AFA-BC1958588160}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SourceGenerators", "SourceGenerators", "{14D18F51-6B59-49D5-9AB7-08B38417A459}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SourceGeneratorSamples", "samples\CSharp\SourceGenerators\SourceGeneratorSamples\SourceGeneratorSamples.csproj", "{2ADE5CFA-5DF4-44A9-BD67-E884BCFBA045}"
+EndProject
```
cartermp, line 117: This appears to be missing the consumption project.

samples/CSharp/SourceGenerators/README.md
chsienki: Yes, it might help if I added that file to the repo huh?

jaredpar:
Why did you need to downgrade the demo? This is now using a non-supported `netcoreapp3.0` version. Is this a problem with the SDK repo? If so we can tackle in a later PR.

chsienki:
Yeah, it wasn't able to build netcoreapp3.1 

Looks like that just pushed the problem further though, because it's now using a compiler that doesn't support SourceGenerators. 

I think I'll just remove them from the main 'shipping' `samples.sln` and keep the 'special' `SourceGenerators.sln` for now. We won't get CI, but we can fix that up in another PR?

jmarolf:
@chsienki what was the original failure about?

chsienki:
@jmarolf https://dev.azure.com/dnceng/public/_build/results?buildId=619533&view=logs&j=a846d25a-e32c-5640-1b53-e815fab94407&t=aafa82ad-f831-5b30-c619-9b12d27e8ee8

jmarolf:
ah change the global.json from this
```json
{
  "tools": {
    "dotnet": "3.0.101",
    "vs": {
      "version": "16.3"
    },
    "xcopy-msbuild": "16.3.0-alpha"
  },
  "msbuild-sdks": {
    "Microsoft.DotNet.Arcade.Sdk": "5.0.0-beta.20117.3"
  }
}
```

to this
```json
{
  "tools": {
    "dotnet": "3.1.101",
    "vs": {
      "version": "16.3"
    },
    "xcopy-msbuild": "16.3.0-alpha"
  },
  "msbuild-sdks": {
    "Microsoft.DotNet.Arcade.Sdk": "5.0.0-beta.20117.3"
  }
}
```

chsienki:
That won't fix it though, because 3.1.101 doesn't include the source generator support. 

So we need to either make the whole repo depend on 3.1.300 or remove them from the main samples solution.

jmarolf:
ah, ok. I am fine with that, There are lots of repos that depend on un-released versions of the SDK

jaredpar:
🎉

PR #527, opened 2020-05-07
Add Csv Generator sample

lucabol:
Reads csv file(s) passed in as `AdditionalFiles` and generates code, which can be called as:

````csharp
foreach(var f in CSV.Foos.All) {
    WriteLine($"{f.Id}, {f.Name}");
}
````
The generated code doesn't need to access the csv file at run time. The objects are created on demand, and added to the `All` static property, which caches them.

````csharp
var c = new Foos();
c.Id = 3
c.Name = "blah"

foos.Add(c);
````

Additional usage loading patterns are not implemented. An idea is to vary the code generation pattern based on a system of attributes related to file/directory:
1. Loading from csv at application startup
2. Loading from csv on demand at first access caching results
3. Loading from csv on demand without caching results

````csharp
[assembly:CsvLoadAtStartup("xxx.csv")]
[assembly:CsvLoadWithoutCache("yyy.csv")]
````


chsienki:
Closing in favor #538

PR #530, opened 2020-05-14
Add more complicated example of CSVGenerator

lucabol:
Creates types from a CSV file and loads objects of those types, so that they are accessible in a strongly typed manner. The CSV files are passed in using assembly attributes. The code to retrieve such csv file names and options is almost certainly wrong, even if it works in the scenarios tested, hence the Draft Pull Request.

An example of usage is:

~~~~csharp
using CSV;

[assembly: CsvFileLoadOptions(LoadTime.Startup, true, @"C:\dev\roslyn-sdk\samples\CSharp\SourceGenerators\GeneratedDemo\Cars.csv")]
[assembly: CsvFileLoadOptions(LoadTime.OnDemand, true, @"C:\dev\roslyn-sdk\samples\CSharp\SourceGenerators\GeneratedDemo\People.csv")]

namespace GeneratedDemo
{
    class UseCsvGenerator
    {
        public static void Run()
        {
           WriteLine("## CARS");
           Cars.All.ToList().ForEach(c => WriteLine($"{c.Brand}\t{c.Model}\t{c.Year}\t{c.Cc}"));
           WriteLine("\n## PEOPLE");
           People.All.ToList().ForEach(p => WriteLine($"{p.Name}\t{p.Address}\t{p._11Age}"));
        }
    }
}

~~~~


chsienki:
Love the samples here, and I think the more complex one is a great example of how to build something more than just a 'toy' with SG.

My one concern is adding all the CSV parser stuff in as well; it's a bunch of code that we don't necessarily want to have to maintain. Does it (or could it?) exist as a NuGet package somewhere that we could just reference here instead?

jmarolf:
@chsienki I guess this is where I point out that source generators do not work with nuget packages and its unclear if that work will get done for 1.0 of source generators

chsienki:
@jmarolf This is true for distribution, but it's not a problem in this case because we control both the generator and the consumption (we're just referencing via project ref)

jmarolf:
Good point, I just worry about folks getting themselves into trouble down the line. But I think this is fine

lucabol:
I have experimented with various CSV parsers. Choose this one because it is API compatible with https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualbasic.fileio.textfieldparser, which is in .net core, but not in .net standard (yet). If it gets into .net standard, then we can simply remove the file. It is also on Nuget, so if we support Nuget in the future, we can remove the file. It is also one of the few csv parsers that is self contained in a single file. I didn't know what else to do.

chsienki:
Closing in favor of #538

PR #538, opened 2020-06-02
Add Mustache generator sample

lucabol:
[Mustache](https://mustache.github.io/) template string generation done at compile time, including support for repeated sections, unescaped elements etc ...

~~~~csharp
[assembly: Mustache("Lottery", t1, h1)]

...

        public const string t1 = @"
Hello {{name}}
You have just won {{value}} dollars!
{{#in_ca}}
Well, {{taxed_value}} dollars, after taxes.
{{/in_ca}}
";
        public const string h1 = @"
{
  ""name"": ""Chris"",
  ""value"": 10000,
  ""taxed_value"": 5000,
  ""in_ca"": true
}
";

...

            WriteLine(Mustache.Constants.Lottery);
~~~~


samples/CSharp/SourceGenerators/GeneratedDemo/GeneratedDemo.csproj
```diff
@@ -6,17 +6,25 @@
     <LangVersion>preview</LangVersion>
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Remove="People.csv" />
+  </ItemGroup>
+
   <ItemGroup>
     <AdditionalFiles Include="MainSettings.xmlsettings" />
+    <Content Include="People.csv">
```
chsienki, line 15: These should probably be input as additional files, rather than being read from the output directory. Currently if you clean and build it fails: there is essentially a deadlock where the build can't succeed because the files are missing, and the files wont be there until the build succeeds.

lucabol: Ok, make sense. BTW: I don't see the build failing but perhaps we are on a different vs version.

samples/CSharp/SourceGenerators/GeneratedDemo/GeneratedDemo.csproj
```diff
@@ -7,16 +7,23 @@
...
   <ItemGroup>
-    <AdditionalFiles Include="MainSettings.xmlsettings" />
-    <None Include="MainSettings.xmlsettings">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </None>
+    <AdditionalFiles Include="People.csv" CsvLoadType="Startup" />
+    <AdditionalFiles Include="Cars.csv" CsvLoadType="OnDemand" CacheObjects="true" />
+
+    <AdditionalFiles Include="MainSettings.xmlsettings" CopyToOutputDirectory="Always" />
+    <None Include="MainSettings.xmlsettings" CopyToOutputDirectory="Always" /> <!-- TODO: remove this when AdditionalFiles supports CopyToOutputDirectory -->
```
sharwell, line 14: ❗ _Never_ use `Always` for this property (it disables incremental builds and all up-to-date checks for the project). Use `PreserveNewest` instead.

samples/CSharp/SourceGenerators/SourceGeneratorSamples/SourceGeneratorSamples.csproj
```diff
@@ -10,8 +10,18 @@
...
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.6.0-3.20207.2" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.0.0-beta2.final" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.7.0-4.20318.4" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- Generator dependencies -->
+    <PackageReference Include="CsvTextFieldParser" Version="1.2.2-preview" GeneratePathProperty="true" PrivateAssets="all" />
+    <None Include="$(PKGCsvTextFieldParser)\lib\netstandard2.0\*.dll" CopyToOutputDirectory="PreserveNewest" Visible="false" />
```
sharwell, line 19: 📝 These should probably be using `CopyLocalLockFileAssemblies` instead

samples/CSharp/SourceGenerators/GeneratedDemo/GeneratedDemo.csproj
```diff
@@ -7,16 +7,23 @@
...
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\SourceGeneratorSamples\SourceGeneratorSamples.csproj" 
-                      OutputItemType="Analyzer"
-                      ReferenceOutputAssembly="false" />
+    <ProjectReference Include="..\SourceGeneratorSamples\SourceGeneratorSamples.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="true" />
```
sharwell, line 18:
```suggestion
    <ProjectReference Include="..\SourceGeneratorSamples\SourceGeneratorSamples.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```

chsienki: We're using the attributes from the assembly, so in this case we need to reference the output assembly too. We could split the dependencies out into generators and runtime assemblies.

sharwell: The previous approach (generating the attribute) seemed like a good approach. Is there a reason we cannot do that?

samples/CSharp/SourceGenerators/SourceGeneratorSamples/CSVGenerator.cs
```diff
@@ -0,0 +1,192 @@
...
+using NotVisualBasic.FileIO;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System.Collections.Immutable;
+
+#nullable enable
+
+// CsvTextFileParser from https://github.com/22222/CsvTextFieldParser adding suppression rules for default VS config
+
+#pragma warning disable IDE0008 // Use explicit type
```
sharwell, line 21: 📝 This suppression should be removed and the diagnostics fixed before this is merged

samples/CSharp/SourceGenerators/SourceGeneratorSamples/MustacheGenerator.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection.Metadata;
+using System.Text;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Text;
+
+#pragma warning disable IDE0008 // Use explicit type
```
sharwell, line 13: 📝 This suppression should be removed and the diagnostics fixed before this is merged

samples/CSharp/SourceGenerators/SourceGeneratorSamples/CSVGenerator.cs
lucabol: Honestly, I wondered why this sample is set up so that you are forced to give explicit types to all variables. This is not the default for VS, not what we recommend and rather cumbersome to write code this way. If it were me, I would change this sample so that one can use 'var' freely. But I didn't want to do it myself as didn't know the reasons for it.

sharwell: > Honestly, I wondered why this sample is set up so that you are forced to give explicit types to all variables.

I'm not sure _why_ it's configured this way, but changing the code style where it's already configured in a strict manner would be a separate pull request.

sharwell: I'm working with @chsienki to clean up the details on this pull request so it can be merged as new examples. We should have more information over the next week.

lucabol: Ok. I was just commenting on how strange the setup was for this sample and wondered if we should fix it at the same time. I don't think anyone would mind us doing it in one PR, but you guys decide.

eng/Versions.props
```diff
@@ -21,6 +21,7 @@
     <MicrosoftCodeAnalysisEditorFeaturesTextVersionVS2019>3.3.1</MicrosoftCodeAnalysisEditorFeaturesTextVersionVS2019>
     <MicrosoftCodeAnalysisVisualBasicVersionVS2019>3.3.1</MicrosoftCodeAnalysisVisualBasicVersionVS2019>
     <MicrosoftCodeAnalysisWorkspacesCommonVersionVS2019>3.3.1</MicrosoftCodeAnalysisWorkspacesCommonVersionVS2019>
+    <MicrosoftNetCompilersToolsetVersion>3.7.0-4.20324.3</MicrosoftNetCompilersToolsetVersion>
```
sharwell, line 24: ❗ This reverted the value from line 13 above

Issue #539, opened 2020-06-02
Fix VSTHRD010 diagnostics


sharwell:
VSTHRD010 is disabled in several locations in code. Each of these should be fixed and the suppression removed.

PR #540, opened 2020-06-02
Update VSSDK dependencies to Visual Studio 16.5

sharwell:
Issue #539 has been filed to address the diagnostics which are now suppressed.


jmarolf:
I am finw with the change in spirit, though I would like to know what this is going to fix.

sharwell:
@jmarolf I needed APIs from dev15 for `ITableEntry` so I can make the integration tests more reliable.

jmarolf:
gotcha

PR #543, opened 2020-06-03
Add regression test for DgmlHelper resources

sharwell:
The test is expected to fail in this pull request, in order to validate that the new unit test project is included in the build.

See #327


sharwell:
The first test pass failed. 🎉  Now that #542 is merged, I'll rebuild this and the tests should pass.

PR #552, opened 2020-06-25
Add ResxSourceGenerator

sharwell:
This is generally extracted from https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/src/GenerateResxSource.cs.


samples/CSharp/SourceGenerators/SourceGeneratorSamples/ResxSourceGenerator.cs
```diff
@@ -0,0 +1,663 @@
...
+                return;
+            }
+
+            if (!context.AnalyzerConfigOptions.GlobalOptions.TryGetValue("build_property.RootNamespace", out string rootNamespace))
+            {
+                rootNamespace = context.Compilation.AssemblyName;
+            }
+
+            string resourceName = Path.GetFileNameWithoutExtension(resourceFile.Path);
+            if (options.TryGetValue("build_metadata.AdditionalFiles.RelativeDir", out string relativeDir))
```
sharwell, line 58: @chsienki This value is appearing in the generated global **.editorconfig** file, but it's not appearing in `options`

chsienki: This was fixed in https://github.com/dotnet/roslyn/pull/45357 but hasn't flowed through to any published packages yet.

sharwell: ➡️ Fixed by #559

sharwell:
Superseded by https://github.com/dotnet/roslyn-analyzers/pull/4897

PR #553, opened 2020-06-25
Save sample source generator output

sharwell:
Generalizes the pattern described in https://github.com/dotnet/roslyn-sdk/issues/526#issuecomment-630770165, and makes the generation a characteristic of the target project instead of the source generator.


chsienki:
We're going to be doing this as part of the actual compilation in the next preview. We probably don't want to check in a hack that will be pretty quickly made obsolete. https://github.com/dotnet/roslyn/projects/54#card-34161504

PR #563, opened 2020-07-15
Code cleanup for CsvGenerator and MustacheGenerator

sharwell:
Review commit-by-commit is recommended.


samples/CSharp/SourceGenerators/GeneratedDemo/GeneratedDemo.csproj
```diff
@@ -18,12 +18,7 @@
     <ProjectReference Include="..\SourceGeneratorSamples\SourceGeneratorSamples.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
   </ItemGroup>
 
-  <!-- Manually reference the generator dependencies and props because we locally reference the generator. When added via NuGet this happens automatically -->
-  <Target AfterTargets="ResolveProjectReferences" Name="GetGeneratorDependencies">
-    <ItemGroup>
-      <Analyzer Include="%(RootDir)%(Directory)\*.dll" Exclude="@(Analyzer)" />
-    </ItemGroup>
-  </Target>
```
sharwell, line 26: 📝 I was expecting removing this to break something and need to be addressed separately, but everything kept working.

eng/Versions.props
```diff
@@ -21,7 +21,6 @@
     <MicrosoftCodeAnalysisEditorFeaturesTextVersionVS2019>3.3.1</MicrosoftCodeAnalysisEditorFeaturesTextVersionVS2019>
     <MicrosoftCodeAnalysisVisualBasicVersionVS2019>3.3.1</MicrosoftCodeAnalysisVisualBasicVersionVS2019>
     <MicrosoftCodeAnalysisWorkspacesCommonVersionVS2019>3.3.1</MicrosoftCodeAnalysisWorkspacesCommonVersionVS2019>
-    <MicrosoftNetCompilersToolsetVersion>3.7.0-4.20324.3</MicrosoftNetCompilersToolsetVersion>
```
sharwell, line 24: 📝 This was already set a few lines above

samples/CSharp/SourceGenerators/GeneratedDemo/GeneratedDemo.csproj
jmarolf: I would expect that `OutputItemType="Analyzer"` from the project reference would add the analyzer and since all its dependencies are next to it the CLR will load them without this target that explicitly adds them

---
In reply to: [455084495](https://github.com/dotnet/roslyn-sdk/pull/563#discussion_r455084495) [](ancestors = 455084495)

chsienki: :( this was needed for IDE support. It doesn't look like it breaks because the IDE doesn't yet have the right reload mechanisms hooked up, but if you restart VS after this change you'll get lots of errors in the error list.

sharwell: I restarted the IDE multiple times as part of my testing for this scenario, and it did not impact behavior.

PR #565, opened 2020-07-15
Restore dependency refs in generated project

chsienki:
Fixes https://github.com/dotnet/roslyn-sdk/pull/563/commits/961dd65aa6a72847aa4fdb5eee67d5e8667bbabd#r455431507


sharwell:
I was able to reproduce the described behavior, but fixed it in a different way in #566. The approach in that PR allows **GeneratedDemo** to avoid the complexity of dealing with the dependencies.

PR #568, opened 2020-07-21
Add arcade publishing step

JoeRobich:
Test build showing publishing stages will run https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3909640&view=results


eng/Signing.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+  </PropertyGroup>
```
sharwell, line 4: 📝 This is no longer necessary (MSBuild provides the feature automatically, with lower overhead)

JoeRobich: It complained with there was no signing.props file. I will remove this property group.

sharwell: Right, I was specifically referring to this property group.

JoeRobich:
```suggestion
```

JoeRobich:
After upgrading Arcade symbol Validate Build assets stage completes - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3931729&view=results

PR #569, opened 2020-07-21
Automatically detect generated code configuration

sharwell:
This change eliminates the need to specify `SkipGeneratedCodeCheck` when the analyzer calls `ConfigureGeneratedCodeAnalysis` explicitly.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerInfo.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerInfo.cs
@@ -1,0 +1,171 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
+using System.Threading;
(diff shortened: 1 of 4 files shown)


THIRD-PARTY-NOTICES.txt
```diff
@@ -0,0 +1,38 @@
...
+distributed under licenses different than the Roslyn SDK software.
+
+In the event that we accidentally failed to list a required notice, please
+bring it to our attention. Post an issue or email us:
+
+           dotnet@microsoft.com
+
+The attached notices are provided for information only.
+
+License notice for StreamJsonRpc
```
jmarolf, line 11: Is this change unrelated to the PR? If so please do this in a separate PR. I do not see streamjson RPC being used anywhere

sharwell: A bunch of the dynamic code in `AnalyzerInfo` was derived from the proxy in StreamJsonRpc. I'm checking with @AArnott to see if we need to include this notice or if it was entirely authored here.

jmarolf: ah, I see it was unclear to me that was the reason

AArnott: The only code I recognize is the `SkipVisibilityChecksFor` and related code. But that code didn't originally come from streamjsonrpc -- it came from vs-mef:

https://github.com/microsoft/vs-streamjsonrpc/blob/d64c14cc238cf05f99c981a0ee3d0de2b168fe1f/src/StreamJsonRpc/SkipClrVisibilityChecks.cs#L4

It was entirely authored within Microsoft. So IMO we don't need to list it under 3rd party notices.

PR #584, opened 2020-08-10
Add diagnostic verification extensibility.

nschuessler:
A small extensibility point to allows verification of generated diagnostics for things not yet supported natively by the test library.
In my case, this is the `Properties` dictionary of the diagnostic.

This small change would allow me to remove custom testing frameworks from my projects.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
@@ -239,6 +239,20 @@ namespace Microsoft.CodeAnalysis.Testing
             await VerifySuppressionDiagnosticsAsync(analyzers, sources, additionalFiles, additionalProjects, additionalMetadataReferences, expected, verifier, cancellationToken).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Checks each of the actual <see cref="Diagnostic"/>s found and compares them with the corresponding
+        /// <see cref="DiagnosticResult"/> based on custom test requirements not yet supported by the test framework.
+        /// </summary>
+        /// <param name="analyzers">The analyzers that have been run on the sources.</param>
+        /// <param name="actual">The <see cref="Diagnostic"/>s found by the compiler after running the analyzer
+        /// on the source code.</param>
+        /// <param name="expected">The <see cref="DiagnosticResult"/>s describing the expected
+        /// diagnostics for the sources.</param>
+        /// <param name="verifier">The verifier to use for test assertions.</param>
+        protected virtual void VerifyDiagnosticCustom(ImmutableArray<DiagnosticAnalyzer> analyzers, Diagnostic actual, DiagnosticResult expected, IVerifier verifier)
+        {
+        }
+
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/roslyn-sdk?pullRequest=584) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/roslyn-sdk?pullRequest=584) before we can accept your contribution.<br/><br/>:x: nschuessler [sign now](https://cla.dotnetfoundation.org/dotnet/roslyn-sdk?pullRequest=584)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/roslyn-sdk?pullRequest=584) it.</sub>

nschuessler:
Tagging @sharwell as he has some context for the change.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -133,8 +133,18 @@ public string TestCode
...
+        /// Gets or sets the reference assemblies to use.
+        /// </summary>
         public ReferenceAssemblies ReferenceAssemblies { get; set; } = ReferenceAssemblies.Default;
 
+        /// <summary>
+        /// Gets or sets an additional verifier for a diagnostic.
+        /// The action compares actual <see cref="Diagnostic"/> and the expected
+        /// <see cref="DiagnosticResult"/> based on custom test requirements not yet supported by the test framework.
+        /// </summary>
+        public Action<ImmutableArray<DiagnosticAnalyzer>, Diagnostic, DiagnosticResult, IVerifier>? DiagnosticVerifier { get; set; }
```
sharwell, line 146: Can we remove the analyzer portion?

```suggestion
        public Action<Diagnostic, DiagnosticResult, IVerifier>? DiagnosticVerifier { get; set; }
```

💭 I'm tempted to go as far as removing the `DiagnosticResult` portion as well.

nschuessler: That means callers cant use the following method for consistency:
```
     protected string FormatVerifierMessage(ImmutableArray<DiagnosticAnalyzer> analyzers, Diagnostic actual, DiagnosticResult expected, string message)
```
Is that ok?

nschuessler: re: `DiagnosticResult` for `expected`.  I'd like to keep it so that it can be a key to a dictionary containing expected data.

sharwell: > That means implementors can't use the following method for consistency

The helper wouldn't produce an actionable error message for this case, so I'm not sure how much it helps.

> I'd like to keep it so that it can be a key to a dictionary containing expected data.

📝 This is already not supported. `DiagnosticResult` is a value type which is not equatable.

Let me think about this a bit. If I can't find an "obviously better" option then the one here seems reasonable.

nschuessler: I do use `Span`s as keys / comparisons in some of my analyzers to determine if I'm matching up the same piece of code.

kendrahavens: @sharwell are we okay to go ahead with Niko's contribution?

sharwell: I am

PR #554, opened 2020-06-26
Fix Symbol View In SyntaxTreeVizualizer

jmarolf:
fixes #504

## Before

image

## After

image

Issue #504, opened 2020-04-09
Syntax Visualizer shows SyntaxNodes but no Symbols


bernd5:
In Visual Studio 2019 and Visual Studio 2019 Preview the Syntax Visualizer shows only the type and kind name of a corresponding Symbol - but the grid is empty.
This worked in Visual Studio 2017.

![grafik](https://user-images.githubusercontent.com/26689688/78899171-483de680-7a75-11ea-8a22-9684fca1a318.png)

bernd5:
Is the code in this repository for VS2019 or just VS2017?

SENya1990:
I have the same problem - it works for VS 2017 but not for VS 2019

SENya1990:
It seems I created a duplicate issue:
https://github.com/dotnet/roslyn-sdk/issues/546

bernd5:
The grid seems to be generated via Reflection.
Perhaps there is some Exception while getting the data. The code in this repository looks fine (just a simple assignment) - but what is wrong?
How to debug it?

bernd5:
Any news on this issue? I can work a bit with VS2017 but it's version does not support for example nullable reference types...
Please fix it or provide us some info that we can solve it...

jmarolf:
I'll take a look this week

bernd5:
Any news on this? Can we help you somehow?

bernd5:
really great :-)

bernd5:
Do you have any news on this? We work a lot with roslyn and need the visualizer to use it effectively.

jzielnik:
Has this fix been released? I have similar issue but due fact that .NET 5 is not supported in VS2017 I cannot use it for these projects.

bernd5:
Yes, it has been fixed. But I'm not sure where/when/how it is published.

bernd5:
The issue was, that all symbols are implemented explictitly - and the View worked via reflection which searched for public properties.

PR #634, opened 2020-10-01
Fix Symbol View In SyntaxTreeVizualizer

bernd5:
See:
- https://github.com/dotnet/roslyn-sdk/pull/554
- https://github.com/dotnet/roslyn-sdk/issues/504

=> Use custom TypeDescriptor to display Symbol-Data


sharwell:
Do you know why a custom TypeDescriptor is needed for displaying `ISymbol`, but is not needed for displaying `IOperation`?

bernd5:
What happens inside of the property grid is a call like this:

````
var props = TypeDescriptor.GetProperties(node, new Attribute[] { BrowsableAttribute.Yes });
```
which calls then (more or less):
```
new ReflectTypeDescriptionProvider().GetProperties(node.GetType())
```

For SyntaxNodes it returns some properties (perhaps because of some attributes). But for symbols it returns nothing.

bernd5:
That makes sense, we have the type of the Symbol implementation not the Symbol itself.

bernd5:
We could use the adapter for all kinds of selected objects, it allow a better control...
For example we can print now "None" instead of an empty string.

bernd5:
do you want to merge it or should I change something?

jmarolf (approved):
:shipit:

jmarolf:
Thanks @bernd5! I'll go ahead and merge this in

PR #654, opened 2020-11-09
[Bug fixde] Fixed bug with implicit JObject convertion in source generator mustache sample.

robersonliou:
In the first sample case of mustache section template in `UseMustacheGenerator.cs`, It would get result: "Hello Chris, You have just won 10000 dollars!".
But it should be "Hello Chris, You have just won 10000 dollars! Well, 10000 dollars, after taxes."
The problem here is the implicit json deserialization, so just add `Dictionary<string, object>` as explict type to get pass.


dibarbet:

PR #682, opened 2020-12-22
Add VisualBasic source generator samples

sharwell:


chsienki:
So, the actual code is fine. 

I do wonder if this makes it *harder* for a customer to understand how to make a generator though. Ideally I think we'd have something like:

- Most samples are C#
- One / two that are both VB / C#
- A sample written in VB itself, applying to VB

sharwell:
> One / two that are both VB / C#

Good candidate for this one is .resx generator.

paul1956:
If C# examples were written is simple syntax that was close to VB, it would not be that big an issue. But there is a recent trend to write C# examples using the latest C# features (records, spans, pattern matching, switch expressions...) that have nothing to do with the example and don't port easily to VB.

jmarolf (approved):

sharwell:
I'm going to focus on resx as a cross-language example.

PR #625, opened 2020-09-14
Update generator samples to latest version of the API

chsienki:

PR #690, opened 2020-12-31
Add a source generator for a Maths language to the samples

lucabol:
The scenario is that a financial/gas/whatever-scientific company wants their mathematicians to write their formulas with a syntax familiar to them and without the risk of introducing bugs. But they also need to run as fast as compiled code (i.e. inline), because that is what their business relies on.

Enter source generators. They now can write:

```
AreaSquare(l)       = pow(l, 2)
AreaRectangle(w, h) = w * h
AreaCircle(r)       = pi * r * r
Quadratic(a, b, c)  = {-b + sqrt(pow(b,2) - 4 * a * c)} / (2 * a)

PointDist(x', x'', y', y'')     = sqrt(pow([x'-x''],2) + pow([y'-y''], 2))
```

I needed to build a lexer / parser for the language instead of simple regex substitution to avoid the possibility of an injection attack by a disgruntled employee. I built it in the simplest possible way for the sake of example.

The language supports:
1. Using prime char in identifiers
2. Using differents parenthesis
3. Calling functions in the Math class lowercase and functions defined before in the file (no other functions allowed)
4. Simpler syntax than C# code (i.e. no types etc...)

This is not completed yet, but done a pull request for the sake of review.

### DIFF
--- a/samples/CSharp/SourceGenerators/SourceGeneratorSamples/MathsGenerator.cs
+++ b/samples/CSharp/SourceGenerators/SourceGeneratorSamples/MathsGenerator.cs
@@ -148,7 +148,7 @@ namespace MathsGenerator {
 
         private readonly static string Preamble = @"
 using static System.Math;
-using static ___MathLibrary___.Formulas;
+using static Maths.FormulaHelpers;
 
 namespace Maths {
 
@@ -212,7 +212,7 @@ namespace Maths {
             TokenType.Identifier    => EmitIdentifier(ctx, token),
             TokenType.Number        => ctx.buffer.Append(token.Value),
             TokenType.Operation     => ctx.buffer.Append(token.Value),
-            TokenType.Sum           => ctx.buffer.Append("___MySum___"),
+            TokenType.Sum           => ctx.buffer.Append("MySum"),
             _                       => Error(token, TokenType.None)
         };
 
@@ -415,8 +415,8 @@ using System.Linq;
 using System;
 using System.Collections.Generic;
 
-namespace ___MathLibrary___ {
- public static partial class Formulas {
+namespace Maths {
+ public static class FormulaHelpers {
 
         public static IEnumerable<double> ConvertToDouble(IEnumerable<int> col)
         {
@@ -424,7 +424,7 @@ namespace ___MathLibrary___ {
                 yield return (double) s;
         }
 
-        public static double ___MySum___(int start, int end, Func<double, double> f) =>
+        public static double MySum(int start, int end, Func<double, double> f) =>
             Enumerable.Sum<double>(ConvertToDouble(Enumerable.Range(start, end - start)), f);
     }
 }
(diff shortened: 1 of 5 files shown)


samples/CSharp/SourceGenerators/SourceGeneratorSamples/MathsGenerator.cs
```diff
@@ -0,0 +1,474 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace ___MathLibrary___ {
+ public static partial class Formulas {
+
+        public static IEnumerable<double> ConvertToDouble(IEnumerable<int> col)
+        {
+            foreach (var s in col)
+                yield return (double)Convert.ChangeType(s, typeof(double));
```
jaredpar, line 424: Is there a reason that you didn't just use `yield return (double)s` here? The conversion from `int` to `double` is non-lossy here, can't see the need for `ChangeType`

samples/CSharp/SourceGenerators/SourceGeneratorSamples/MathsGenerator.cs
```diff
@@ -0,0 +1,474 @@
...
+    public class MathsGenerator : ISourceGenerator
+    {
+        private bool libraryIsAdded = false;
+
+        private const string libraryCode = @"
+using System.Linq;
+using System;
+using System.Collections.Generic;
+
+namespace ___MathLibrary___ {
```
jaredpar, line 418: Hmm, I think I get what you're trying to do here (avoid conflicting with user defined types) and yeah we don't have hygenic macros / names in the language. Same time the `__` in the name throws me off a bit. 

Is there a reason you just didn't make it a util style class, like `FormulaHelpers`, inside the `Math` namepace that you're generating?

jaredpar (approved):

samples/CSharp/SourceGenerators/SourceGeneratorSamples/MathsGenerator.cs
lucabol: Fixed in 8fbc804e9af50d2f6bd4b8f5b24962f236386853 . Not sure why I wrote it that way originally.

samples/CSharp/SourceGenerators/SourceGeneratorSamples/MathsGenerator.cs
lucabol: Fixed in 425941d959a4c69588f79e7407841ab1e97fba0a. Great point about hygienic macros. I didn't realize the similarity. I need to remember to touch on that in the blog post.

jaredpar (approved):

jmarolf:
@lucabol can you pull in the latest changes from master? `SourceGeneratorContext` no longer exists in the public API and all of our samples where updated here: https://github.com/dotnet/roslyn-sdk/pull/625

lucabol:
I am closing this PR as it was created from an old fork. #695  represents the new one we are merging.

PR #692, opened 2021-01-06
Added VB (specific/focused) SourceGenerators Sample (Try #2)

DualBrain:
Couldn't figure out how to "rebase" in order to *fix* the last submission; so I started from the beginning.  This one shouldn't have any of the bin/obj folders.

Reference rest of conversation in https://github.com/dotnet/roslyn-sdk/pull/691

Thanks for your consideration.

### DIFF
--- a/samples/VisualBasic/SourceGenerators/GeneratedDemo/Cars.csv
+++ b/samples/VisualBasic/SourceGenerators/GeneratedDemo/Cars.csv
@@ -1,0 +1,4 @@
+﻿Brand, Model, Year, cc, Favorite
+Fiat, Punto, 2008, 12.3, No
+Ford, Wagon, 1956, 20.3, No
+BMW, "335", 2014, 20.3, Yes
--- a/samples/VisualBasic/SourceGenerators/GeneratedDemo/GeneratedDemo.vbproj
+++ b/samples/VisualBasic/SourceGenerators/GeneratedDemo/GeneratedDemo.vbproj
@@ -1,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <AdditionalFiles Include="People.csv" CsvLoadType="Startup" />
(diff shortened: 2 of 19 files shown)


DualBrain:
Forgot to tag @KathleenDollard when I originally submitted this one.  Doing so now.

DualBrain:
If not accepted, would it be possible to add a link to an external repo where the VB'ified versions could be found - thus removing the responsibility by the Roslyn team to maintain a VB specific set of examples.  This link would, of course, have any and all disclaimers and mention that it's external of Microsoft yada, yada, yada...

I ask this as it was suggested that there isn't any reason to have a C# and VB version of the samples as VB developers could *simply* look at the existing C# ones to "get the gist".  I somewhat disagree with this sentiment as I think many VB devs would just assume that it doesn't work with VB since there aren't any examples in it and not *waste their time* investigating it any further (again, lacking examples demonstrating that it does indeed work).  I'm a fan of Source Generators... even if there is clearly more work to be done in some areas... but a fan nonetheless.  So I'd love for members of my community (VB) to share in my excitement.  I have no issue with my version of the samples being *somewhere else*; however, that doesn't stop people from clicking on blog posts that take them to this repo and them seeing nothing for them.  So at least a link pointing them somewhere else would at least something...

I currently have them available as part of my general Samples repo at https://github.com/DualBrain/Samples/tree/master/SourceGenerators

In addition to the existing C# inspired originals, my repo also includes an example that *attempts* to reproduce many of the behaviors of C# Records in VB through a Source Generator; and I'm currently exploring/experimenting with another Source Generator that would work to handle Implicit Interface handling (similar to C#... where possible) that will be added to that repo.

@KathleenDollard

jmarolf:
Looks good @DualBrain to merge this you will need to add the new project to the Samples.sln solution [here](https://github.com/dotnet/roslyn-sdk/blob/main/Samples.sln). Once CI passes with those changes I think we will be good to go.

jmarolf:
> If not accepted, would it be possible to add a link to an external repo where the VB'ified versions could be found - thus removing the responsibility by the Roslyn team to maintain a VB specific set of examples. This link would, of course, have any and all disclaimers and mention that it's external of Microsoft yada, yada, yada...

This argument doesn't apply here. The Roslyn uses both C# and VB on a regular basis and we can certainly understand and own these samples.

jmarolf:
alright, I am going to merge this and fix up the sln file in a different PR

PR #695, opened 2021-01-14
Add a source generator for a Maths language to the samples

lucabol:
The scenario is that a financial/gas/whatever-scientific company wants their mathematicians to write their formulas with a syntax familiar to them and without the risk of introducing bugs. But they also need to run as fast as compiled code (i.e. inline), because that is what their business relies on.

Enter source generators. They now can write:

AreaSquare(l)       = pow(l, 2)
AreaRectangle(w, h) = w * h
AreaCircle(r)       = pi * r * r
Quadratic(a, b, c)  = {-b + sqrt(pow(b,2) - 4 * a * c)} / (2 * a)

PointDist(x', x'', y', y'')     = sqrt(pow([x'-x''],2) + pow([y'-y''], 2))

I needed to build a lexer / parser for the language instead of simple regex substitution to avoid the possibility of an injection attack by a disgruntled employee. I built it in the simplest possible way for the sake of example.

The language supports:

    Using prime char in identifiers
    Using differents parenthesis
    Calling functions in the Math class lowercase and functions defined before in the file (no other functions allowed)
    Simpler syntax than C# code (i.e. no types etc...)
    Summation operator symbol

### DIFF
--- a/samples/CSharp/SourceGenerators/GeneratedDemo/GeneratedDemo.csproj
+++ b/samples/CSharp/SourceGenerators/GeneratedDemo/GeneratedDemo.csproj
@@ -9,6 +9,7 @@
   <ItemGroup>
     <AdditionalFiles Include="People.csv" CsvLoadType="Startup" />
     <AdditionalFiles Include="Cars.csv" CsvLoadType="OnDemand" CacheObjects="true" />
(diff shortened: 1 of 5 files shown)


jmarolf (approved):

jmarolf:
Thanks for this!

PR #726, opened 2021-02-09
Add component debugger

chsienki:
Adds the prototype Roslyn component debugger to the SDK.

## UX

User adds `<IsRoslynComponent>true</IsRoslynComponent>` to their generator project file

This enables the 'Roslyn Component' debugger option in debug settings (Project Properties -> Debug)

image

A user chooses the appropriate target project

image

User F5's the generator project which starts CSC.exe/VBC.exe building the target project allowing them to debug the generator code.

## Known Issues

- Project filtering was hanging on some projects, so is disabled for now. The UI just displays all available projects as targets.
- Localization support is missing
- Target framework support when multi-targeting. Can we add another option?

## FAQ

1. How will this work with VSCode?
    - VSCode supports launchsettings.json so we should be able to write an equivalent VSCode extension that reads the target and does the launching

### DIFF
--- a/src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
+++ b/src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
@@ -14,11 +14,14 @@ namespace Roslyn.ComponentDebugger
     [AppliesTo(ProjectCapabilities.CSharp + " | " + ProjectCapabilities.VB)]
     public class CapabilityProvider : ConfiguredProjectCapabilitiesProviderBase
     {
+        private readonly IProjectSnapshotService snapshotService;
+
         [ImportingConstructor]
         [System.Obsolete("This exported object must be obtained through the MEF export provider.", error: true)]
-        public CapabilityProvider(ConfiguredProject configuredProject)
+        public CapabilityProvider(ConfiguredProject configuredProject, IProjectSnapshotService snapshotService)
             : base(nameof(CapabilityProvider), configuredProject)
         {
+            this.snapshotService = snapshotService;
         }
 
         protected override async Task<ImmutableHashSet<string>> GetCapabilitiesAsync(CancellationToken cancellationToken)
@@ -26,22 +29,15 @@ namespace Roslyn.ComponentDebugger
             // an alternative design could be to have 'IsRoslynComponent' just define the <Capability... directly in the managed.core targets
             // but that would require a specific roslyn version to work, this allows it to be backwards compatible with older SDKs
             var caps = Empty.CapabilitiesSet;
-            if (await IsRoslynComponentAsync(this.ConfiguredProject, cancellationToken).ConfigureAwait(false))
+
+            var snapshot = await snapshotService.GetLatestVersionAsync(ConfiguredProject, cancellationToken: cancellationToken).ConfigureAwait(false);
+            var isRoslynComponentProperty = snapshot.Value.ProjectInstance.GetPropertyValue(Constants.RoslynComponentPropertyName);
+            var isComponent = string.Compare(isRoslynComponentProperty.Trim(), "true", System.StringComparison.OrdinalIgnoreCase) == 0;
+            if (isComponent)
             {
                 caps = caps.Add(Constants.RoslynComponentCapability);
             }
             return caps;
         }
-
-        private static Task<bool> IsRoslynComponentAsync(ConfiguredProject configuredProject, CancellationToken token = default) 
-            => configuredProject.Services.ProjectLockService.ReadLockAsync(
-                    async access =>
-                    {
-                        var project = await access.GetProjectAsync(configuredProject).ConfigureAwait(false);
-                        var isRoslynComponentProperty = project.GetProperty(Constants.RoslynComponentPropertyName);
-                        var isComponent = string.Compare(isRoslynComponentProperty?.EvaluatedValue.Trim(), "true", System.StringComparison.OrdinalIgnoreCase) == 0;
-                        return isComponent;
-                    },
-                    token);
     }
 }
--- a/src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
+++ b/src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
@@ -23,7 +23,8 @@ namespace Roslyn.ComponentDebugger
 
         protected override async Task<ImmutableHashSet<string>> GetCapabilitiesAsync(CancellationToken cancellationToken)
         {
-            // PROTOTYPE: an alternative could be to have 'IsRoslynComponent' just define the <Capability... directly in the managed.core targets?
+            // an alternative design could be to have 'IsRoslynComponent' just define the <Capability... directly in the managed.core targets
+            // but that would require a specific roslyn version to work, this allows it to be backwards compatible with older SDKs
             var caps = Empty.CapabilitiesSet;
             if (await IsRoslynComponentAsync(this.ConfiguredProject, cancellationToken).ConfigureAwait(false))
             {
--- a/src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
+++ b/src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
@@ -11,7 +11,7 @@ using Microsoft.VisualStudio.ProjectSystem;
 namespace Roslyn.ComponentDebugger
 {
     [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectCapabilitiesProvider))]
-    [AppliesTo(ProjectCapabilities.CSharp)]
+    [AppliesTo(ProjectCapabilities.CSharp + " | " + ProjectCapabilities.VB)]
     public class CapabilityProvider : ConfiguredProjectCapabilitiesProviderBase
     {
         [ImportingConstructor]
--- a/src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
+++ b/src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
@@ -1,0 +1,46 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem;
+
+namespace Roslyn.ComponentDebugger
+{
+    [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectCapabilitiesProvider))]
+    [AppliesTo(ProjectCapabilities.CSharp)]
+    public class CapabilityProvider : ConfiguredProjectCapabilitiesProviderBase
+    {
+        [ImportingConstructor]
+        [System.Obsolete("This exported object must be obtained through the MEF export provider.", error: true)]
+        public CapabilityProvider(ConfiguredProject configuredProject)
+            : base(nameof(CapabilityProvider), configuredProject)
+        {
+        }
+
+        protected override async Task<ImmutableHashSet<string>> GetCapabilitiesAsync(CancellationToken cancellationToken)
+        {
+            // PROTOTYPE: an alternative could be to have 'IsRoslynComponent' just define the <Capability... directly in the managed.core targets?
+            var caps = Empty.CapabilitiesSet;
+            if (await IsRoslynComponentAsync(this.ConfiguredProject, cancellationToken).ConfigureAwait(false))
+            {
+                caps = caps.Add(Constants.RoslynComponentCapability);
+            }
+            return caps;
+        }
+
+        private static Task<bool> IsRoslynComponentAsync(ConfiguredProject configuredProject, CancellationToken token = default) 
+            => configuredProject.Services.ProjectLockService.ReadLockAsync(
+                    async access =>
+                    {
+                        var project = await access.GetProjectAsync(configuredProject).ConfigureAwait(false);
+                        var isRoslynComponentProperty = project.GetProperty(Constants.RoslynComponentPropertyName);
+                        var isComponent = string.Compare(isRoslynComponentProperty?.EvaluatedValue.Trim(), "true", System.StringComparison.OrdinalIgnoreCase) == 0;
+                        return isComponent;
+                    },
+                    token);
+    }
+}
--- a/src/VisualStudio.Roslyn.SDK/ComponentDebugger/Constants.cs
+++ b/src/VisualStudio.Roslyn.SDK/ComponentDebugger/Constants.cs
@@ -4,7 +4,7 @@
 
 namespace Roslyn.ComponentDebugger
 {
-    internal class Constants
+    internal static class Constants
     {
         public const string RoslynComponentPropertyName = "IsRoslynComponent";
 
--- a/src/VisualStudio.Roslyn.SDK/ComponentDebugger/Constants.cs
+++ b/src/VisualStudio.Roslyn.SDK/ComponentDebugger/Constants.cs
@@ -1,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Roslyn.ComponentDebugger
+{
+    internal class Constants
+    {
+        public const string RoslynComponentPropertyName = "IsRoslynComponent";
+
+        public const string RoslynComponentCapability = "RoslynComponent";
+
+        public const string CommandName = "DebugRoslynComponent";
+
+        public const string TargetProjectPropertyName = "targetProject";
+    }
+}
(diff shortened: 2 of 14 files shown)


src/VisualStudio.Roslyn.SDK/ComponentDebugger/Roslyn.ComponentDebugger.csproj
```diff
@@ -0,0 +1,35 @@
...
+    <Reference Include="System.Design" />
+    <Reference Include="System.Drawing" />
+    <Reference Include="System.Windows.Forms" />
+    <Reference Include="System.Xml" />
+    <Reference Include="WindowsBase" />
+    <Reference Include="PresentationCore" />
+    <Reference Include="PresentationFramework" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.VisualStudio.ProjectSystem.Managed" Version="16.8.1-beta1-928-02" PrivateAssets="all" />
```
chsienki, line 24: ⁉ These packages don't seem to be on a public feed anywhere? It works when you add one of the devdiv feeds, but they are not public. I tried to get it working with earlier packages but kept hitting NuGet versioning hell. We'll need to fix this for release, but I don't have a great idea how to.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+                }
+            }
+
+            //PROTOTYPE: we probably shouldn't return anything when we couldn't figure it out
+            return new IDebugLaunchSettings[] { settings };
+        }
+
+        private static string GetCompilerRoot(SVsServiceProvider? serviceProvider)
+        {
+            // PROTOTYPE: we should try and work out the compiler location from the project itself
```
chsienki, line 77: ❓ Today this is finding the .NET Framework CSC/VBC via a known path. Is there a way we can query the MSBuild project to figure out the compiler we should use (including Toolset packages etc?)

src/VisualStudio.Roslyn.SDK/ComponentDebugger/ProjectUtilities.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.VisualStudio.ProjectSystem;
+
+namespace Roslyn.ComponentDebugger
+{
+    public static class ProjectUtilities
+    {
+        // PROTOTYPE: is there a way to get this other than hardcoding it?
+        static readonly string[] CommandLineSchemaRuleNames = new[] { "CompilerCommandLineArgs" };
+
+        public static async Task<IList<string>> GetCompilationArgumentsAsync(this ConfiguredProject project)
```
chsienki, line 19: 📝This is a place I'd like to get the project system folks to take a look at how we're getting the value.

chsienki:
@mikadumont @jaredpar @jasonmalinowski @sharwell for initial review of this. The code is in a *fairly* good shape, but there are `PROTOTYPE` comments that we'll want to address before shipping.

I'd also like to get the project system folks involved as it's based on their stuff. @jjmew Can we get some of your folks to take a look at the implementation?

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Constants.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Roslyn.ComponentDebugger
+{
+    internal class Constants
```
jaredpar, line 7:
```suggestion
    internal static class Constants
```

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
jaredpar: I don't think there is a better way right now.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+            var shell = (IVsShell?)serviceProvider?.GetService(typeof(SVsShell));
+            shell?.GetProperty((int)__VSSPROPID2.VSSPROPID_InstallRootDir, out rootDir);
+            return Path.Combine((string)rootDir, "MSBuild", "Current", "Bin", "Roslyn");
+        }
+
+        private static bool TryGetTargetProject(ConfiguredProject project, ILaunchProfile? profile, out UnconfiguredProject? targetProject)
+        {
+            var targetProjectPath = profile?.OtherSettings?.ContainsKey(Constants.TargetProjectPropertyName) == true
+                                    ? profile.OtherSettings[Constants.TargetProjectPropertyName].ToString()
+                                    : string.Empty;
```
jaredpar, line 88: Hmm, is there some variant of `TryGetValue` on `OtherSettings` here?

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+
+                // try and get the configured version of the target project
+                var targetProject = await targetProjectUnconfigured.GetSuggestedConfiguredProjectAsync().ConfigureAwait(false);
+                if (targetProject is object)
+                {
+                    // get its compilation args
+                    var args = await targetProject.GetCompilationArgumentsAsync().ConfigureAwait(false);
+
+                    // append the command line args to the debugger launch
+                    settings.Arguments = string.Join(" ", args);
```
jaredpar, line 67: Curious: if there are say spaces in soucrce file paths will the `args` local already contain `"` around them?

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
chsienki: No, but I added another method to make the logic clearer

---
In reply to: [573354575](https://github.com/dotnet/roslyn-sdk/pull/726#discussion_r573354575) [](ancestors = 573354575)

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
chsienki: Yes, each argument is correctly quoted (and I double checked with a test run)

---
In reply to: [573355409](https://github.com/dotnet/roslyn-sdk/pull/726#discussion_r573355409) [](ancestors = 573355409)

jaredpar (approved):
Modulo getting the prototype comments shifted to point to an issue that we can trakc

chsienki:
Created #728, #729, #730, #731 to track remaining prototype comments

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Roslyn.ComponentDebugger.csproj
jmarolf: @dmonroym what it the correct way to get the latest packages?

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+        private readonly string _compilerRoot;
+
+        [ImportingConstructor]
+        [Obsolete("This exported object must be obtained through the MEF export provider.", error: true)]
+        public DebugProfileProvider(ConfiguredProject configuredProject, IDebugTokenReplacer tokenReplacer, SVsServiceProvider? serviceProvider)
+        {
+            _configuredProject = configuredProject;
+            _tokenReplacer = tokenReplacer;
+            _compilerRoot = GetCompilerRoot(serviceProvider);
+        }
```
davkean, line 36: You can't call a VS service in a MEF component, this will deadlock if someone is blocked on this component on the UI thread and this is created off UI thread. Instead, wrap in an AsyncLazy that switches to the UI thread before getting the value.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Roslyn.ComponentDebugger.csproj
dmonroym: It's published in https://dev.azure.com/azure-public/vside/_packaging?_a=feed&feed=vs-impl

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+            : base(nameof(CapabilityProvider), configuredProject)
+        {
+        }
+
+        protected override async Task<ImmutableHashSet<string>> GetCapabilitiesAsync(CancellationToken cancellationToken)
+        {
+            // an alternative design could be to have 'IsRoslynComponent' just define the <Capability... directly in the managed.core targets
+            // but that would require a specific roslyn version to work, this allows it to be backwards compatible with older SDKs
+            var caps = Empty.CapabilitiesSet;
+            if (await IsRoslynComponentAsync(this.ConfiguredProject, cancellationToken).ConfigureAwait(false))
```
davkean, line 29: This is mixing dataflow and live project data, which you can't do.

You need to listen to and pull on the ProjectInstance via dataflow, here's an example: https://github.com/dotnet/project-system/blob/b7fd4c439d1d39b656d89a8b88464fabffcc8e59/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Web/ProjectTypeGuidsDataSource.cs.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Roslyn.ComponentDebugger.csproj
jmarolf: thanks @chsienki this feed is public, we should just add it to the nuget.config for this repo

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
chsienki: @davkean Oh, interesting. There are classes in the devdiv repo that use this exact pattern. The base class seems to subscribe to events, so I figured it was designed to explicitly not need dataflow stuff. 

I'll try and use the example to create a dataflow version.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
chsienki: Apparently I just didn't find it before. There is. And i'm now using it.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+            _configuredProject = configuredProject;
+            _tokenReplacer = tokenReplacer;
+            _compilerRoot = new AsyncLazy<string>(() => GetCompilerRootAsync(serviceProvider), ThreadHelper.JoinableTaskFactory);
+        }
+
+        public Task OnAfterLaunchAsync(DebugLaunchOptions launchOptions, ILaunchProfile profile) => Task.CompletedTask;
+
+        public Task OnBeforeLaunchAsync(DebugLaunchOptions launchOptions, ILaunchProfile profile) => Task.CompletedTask;
+
+        public bool SupportsProfile(ILaunchProfile? profile) => Constants.CommandName.Equals(profile?.CommandName, StringComparison.OrdinalIgnoreCase);
```
tmeschter, line 43: You should use `StringComparison.Ordinal` for debug command names.

tmeschter (commented):
Other than working out the data flow issue, mostly looks good.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
```diff
@@ -14,34 +14,30 @@ namespace Roslyn.ComponentDebugger
...
         }
 
         protected override async Task<ImmutableHashSet<string>> GetCapabilitiesAsync(CancellationToken cancellationToken)
         {
             // an alternative design could be to have 'IsRoslynComponent' just define the <Capability... directly in the managed.core targets
             // but that would require a specific roslyn version to work, this allows it to be backwards compatible with older SDKs
             var caps = Empty.CapabilitiesSet;
-            if (await IsRoslynComponentAsync(this.ConfiguredProject, cancellationToken).ConfigureAwait(false))
+
+            var snapshot = await snapshotService.GetLatestVersionAsync(ConfiguredProject, cancellationToken: cancellationToken).ConfigureAwait(false);
```
chsienki, line 33: Thanks @davkean and @tmeschter. I switched the cap provider over to this, which is the same way the 'get command line args' stuff works. Is that a valid way to handle getting it via dataflow?

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Immutable;
+using System.ComponentModel.Composition;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.ProjectSystem;
+
+namespace Roslyn.ComponentDebugger
+{
+    [Export(ExportContractNames.Scopes.ConfiguredProject, typeof(IProjectCapabilitiesProvider))]
+    [AppliesTo(ProjectCapabilities.CSharp + " | " + ProjectCapabilities.VB)]
```
davkean, line 14: This component will load for every C# and VB project, we need a better approach for this? Is there something unique about these projects?

davkean:
This got merged in without any sign off from any project-system folks, can we please undo that?

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
davkean: This also loads for Shared Projects.

chsienki: It can apply to any project that has the `<IsRoslyComponent>true</IsRoslynComponent>`. So we load the provider for all projects and add the capability to any that match that. 

Is there some better way to achieve that? I mentioned in the comments we could do it via the targets, but that does require a matching roslyn version.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
davkean: No. This needs to chain similar to what I pointed out here: https://github.com/dotnet/project-system/blob/feature/AspNet/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/Web/ProjectTypeGuidsDataSource.cs. 

These systems flow data from one data source -> another, you need to transform the data as it flows along the pipe. This code just grabs the latest version from the snapshot pipe and there's sets up no relationship between the pipes.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
davkean: Is there a targets file that it imports we can match on?

chsienki: No, its just a regular old csharp project.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
chsienki: But isn't that handled by the base class? It sets up the pipe stuff and calls this method as it changes? I guess thats the bit I don't understand.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
davkean: > I mentioned in the comments we could do it via the targets, but that does require a matching roslyn version.

Can you expand on what you mean by this? You mean if the project itself has a package reference to a different version of the compiler?

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
davkean: No. Here you basically saying "I am the start of a pipe, and I will tell you when there are changes". What you want to say is "My pipe is based on the snapshot data, so when it changes, call me".

davkean: This stuff is complicated and hard to get right - I can sit down and write it with you.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CapabilityProvider.cs
chsienki: Yeah. We could drop the capability provider altogether if we just added the following to the targets file

```xml
<ItemGroup Condition="'$(IsRoslynComponent)' == 'true'">
    <Capability Include="RoslynComponent">
</ItemGroup>
```

But anyone using an older version of the compiler via NuGet wouldn't get the capability, and be able to use it. That might be worth the tradeoff to remove some of the complexity here tho.

jmarolf:
I've reverted this change, sorry @chsienki! I think you will need to submit a new one.

PR #734, opened 2021-02-23
Add support for .editorconfig documents

sharwell:


jmarolf:
Thanks @sharwell! Lets merge this as soon as CI passes

tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.UnitTests/AnalyzerConfigFilesTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            await new CSharpTest
+            {
+                TestState =
+                {
+                    Sources = { "namespace MyNamespace { }" },
+                    ExpectedDiagnostics = { new DiagnosticResult(HighlightBracesIfAnalyzerConfigMissingAnalyzer.Descriptor).WithLocation(1, 23) },
+                    AnalyzerConfigFiles =
+                    {
+                        ("/.editorconfig", SourceText.From(RootEditorConfig, Encoding.UTF8)),
+                    },
```
Youssef1313, line 41: Wondering when this can be different than using AdditionalFiles like https://github.com/dotnet/roslyn-analyzers/blob/1ead3c96987ec6200209ab602529b9bab573833d/src/NetAnalyzers/UnitTests/Microsoft.CodeQuality.Analyzers/ApiDesignGuidelines/DoNotPrefixEnumValuesWithTypeNameTests.cs#L172 for example?

sharwell: > Wondering when this can be different than using AdditionalFiles

An **.editorconfig** which is added only as an additional file is meaningless on its own. The only reason that analyzer works is it [manually parses the file](https://github.com/dotnet/roslyn-analyzers/blob/3fb878177fb161344a57784957485450544e0f84/src/Utilities/Compiler/Options/AnalyzerOptionsExtensions.cs#L662-L678).

Youssef1313: @shawell Ahhhh, that's why it works in roslyn-analyzers tests without a `[*.cs]` section while the compiler requires it.

jmarolf (approved):

PR #738, opened 2021-02-24
Component debugger

chsienki:
Follows on from #726.

 - Removes the `CapabilityProvider` and we'll set it through the roslyn targets instead. I'll issue a follow up PR on the roslyn side
 - Reworks the command line arg obtaining path to use dataflow instead
 - Refactors the view model and centralizes the launch profile reading, so relative paths + msbuild tokens work in all places


chsienki:
@davkean here's the cleaned up implementation based on yesterdays work. It's still using the 'private' rule name, but otherwise I think is ready to go. Would you mind casting your eye over it again to make sure it's still doing things correctly?

chsienki:
@jaredpar @jmarolf for review.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.VisualStudio.ProjectSystem.VS.Debug;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Threading;
+using Task = System.Threading.Tasks.Task;
+
+namespace Roslyn.ComponentDebugger
+{
+    [Export(typeof(IDebugProfileLaunchTargetsProvider))]
+    [AppliesTo(Constants.RoslynComponentCapability)]
```
jmarolf, line 23: Do we want to update the projects in the templates to have this capability?

jmarolf (approved):
:shipit:

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        private async Task<string> GetCompilerRootAsync(SVsServiceProvider? serviceProvider)
+        {
+            await _threadingService.SwitchToUIThread();
+
+            // https://github.com/dotnet/roslyn-sdk/issues/729 : don't hardcode net fx compiler
+            object rootDir = string.Empty;
+            var shell = (IVsShell?)serviceProvider?.GetService(typeof(SVsShell));
+            shell?.GetProperty((int)__VSSPROPID2.VSSPROPID_InstallRootDir, out rootDir);
+            return Path.Combine((string)rootDir, "MSBuild", "Current", "Bin", "Roslyn");
```
jaredpar, line 86: Feel like the `?.` is hiding a bug here. If `shell is null` here then we end up calling `Path.Combine("", ...)` which will just return a relative path to `csc.exe` vs a full one. That's not going to execute the way we like. Think we should just proactively error here if `shell is null`

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebuggerOptionsViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public int SelectedProjectIndex
+        {
+            get => _selectedProjectIndex;
+            set
+            {
+                _selectedProjectIndex = value;
+                NotifyPropertyChanged();
+                indexChanged(value);
```
jaredpar, line 45: Think we should only do these lines if `_selectedProjectIndex` actually changed.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebuggerOptionsViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+            this.indexChanged = indexChanged;
+        }
+
+        public IEnumerable<string> ProjectNames
+        {
+            get => _projectNames;
+            set
+            {
+                _projectNames = value;
+                NotifyPropertyChanged();
```
jaredpar, line 34: Think we should only do this if `_projectNames` changed. Think `SequenceEquals` would work here for the check. Doesn't need to be exhaustive but enough to avoid event storms.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebuggerOptionsViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+
+namespace Roslyn.ComponentDebugger
+{
+    internal class DebuggerOptionsViewModel : INotifyPropertyChanged
```
jaredpar, line 13:
```suggestion
    internal sealed class DebuggerOptionsViewModel : INotifyPropertyChanged
```

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebuggerOptionsViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        private readonly Action<int> indexChanged;
```
jaredpar, line 15:
```suggestion
        private readonly Action<int> _indexChanged;
```

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebuggerOptionsViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        private IEnumerable<string> _projectNames = ImmutableArray<string>.Empty;
+
+        private int _selectedProjectIndex = -1;
+
+        public event PropertyChangedEventHandler? PropertyChanged;
+
+        public DebuggerOptionsViewModel(Action<int> indexChanged)
+        {
+            this.indexChanged = indexChanged;
```
jaredpar, line 25: Should add `Debug.Assert(indexChanged is object)` or only conditionally invoke it later on.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/LaunchSettingsProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+            var targetArg = projectArgs.LastOrDefault(a => a.StartsWith("/out:", StringComparison.OrdinalIgnoreCase));
+            var target = Path.GetFileName(targetArg);
+
+            var projectService = _unconfiguredProject.Services.ProjectService;
+            foreach (var targetProjectUnconfigured in projectService.LoadedUnconfiguredProjects)
+            {
+                // check if the args contain the project as an analyzer ref
+                foreach (var arg in await targetProjectUnconfigured.GetCompilationArgumentsAsync().ConfigureAwait(false))
+                {
+                    if (arg.StartsWith("/analyzer", StringComparison.OrdinalIgnoreCase)
```
jaredpar, line 76:
```suggestion
                    if (arg.StartsWith("/analyzer:", StringComparison.OrdinalIgnoreCase)
```
To be consistent with the `StartsWith` above.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebugProfileProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                    settings.Arguments = string.Join(" ", args);
+                }
+            }
+            // https://github.com/dotnet/roslyn-sdk/issues/728 : better error handling
+            return new IDebugLaunchSettings[] { settings };
+        }
+
+        private async Task<string?> GetCompilerRootAsync(SVsServiceProvider? serviceProvider)
+        {
+            await _threadingService.SwitchToUIThread();
```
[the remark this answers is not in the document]
chsienki: The explicit point of this is to change the captured context: `.ConfigureAwait(true)` would be incorrect, `.ConfigureAwait(false)` would be redundant, so the analyzer doesn't complain about it.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/DebuggerOptions.xaml
```diff
@@ -0,0 +1,19 @@
...
+    <Grid Margin="0,2,0,0">
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="158" />
+            <ColumnDefinition Width="350" />
+            <ColumnDefinition Width="Auto" />
+        </Grid.ColumnDefinitions>
+
+        <!-- https://github.com/dotnet/roslyn-sdk/issues/730 : Localization -->
+        <Label Margin="4,4,3,5">Target Project:</Label>
+        <ComboBox Grid.Column="1" Margin="5,7,2,6" ItemsSource="{Binding ProjectNames}" SelectedIndex="{Binding SelectedProjectIndex, Mode=TwoWay}" />
```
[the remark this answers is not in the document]
chsienki: Haha, I used to be paid to write windows phone apps... sooo.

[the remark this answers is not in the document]
jaredpar: Oh wow, I was joking ... but you are an expert. Now I know who to assign all our WPF bugs too ... ;)

src/VisualStudio.Roslyn.SDK/ComponentDebugger/CommandLineArgumentsDataSource.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Roslyn.ComponentDebugger
+{
+    [Export]
+    [AppliesTo("(" + ProjectCapabilities.CSharp + " | " + ProjectCapabilities.VB + ") & !" + ProjectCapabilities.SharedAssetsProject)]
+    public class CommandLineArgumentsDataSource : UnconfiguredProjectHostBridge<IProjectVersionedValue<IProjectSubscriptionUpdate>, IProjectVersionedValue<ImmutableArray<string>>, IProjectVersionedValue<ImmutableArray<string>>>
+    {
+        private readonly IActiveConfiguredProjectSubscriptionService activeProjectSubscriptionService;
+
+        [ImportingConstructor]
+        public CommandLineArgumentsDataSource(IProjectThreadingService projectThreadingService, IActiveConfiguredProjectSubscriptionService activeProjectSubscriptionService)
```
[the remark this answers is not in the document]
chsienki: Hmm, its not. the `?` is just to shut up the analyzer which doesn't seem to be nullable aware :/

I'll explicitly suppress it though, to make it clear.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Constants.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Roslyn.ComponentDebugger
+{
+    internal static class Constants
+    {
+        public const string RoslynComponentCapability = "RoslynComponent";
+
+        public const string CommandName = "DebugRoslynComponent";
+
+        public const string TargetProjectPropertyName = "targetProject";
```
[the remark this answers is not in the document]
chsienki: No. Its the name of the property in the json launch settings file, which should be camelCased. I'll try and think of a clearer name.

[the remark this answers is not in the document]
jaredpar: Ah okay, yeah camelCase is the only answer then.

jaredpar (approved):

Issue #601, opened 2020-08-31
Analyzer template not working for accounts with long user names


MichaRotstein:
I just tried again to create C# analyzer codefix project with the new Version 16.8.0 Preview 2.0
I get:

![image](https://user-images.githubusercontent.com/65199232/91680247-f87bcb00-eb8d-11ea-8007-d2ba616a399f.png)

I have now been trying to simply create and build the c# analyzer template which should be trivial for over a month with no luck. Many issues have been opened and get closed without this resolved.

sharwell:
I believe this is caused by the path being too long. Can you tell me the total number of characters appearing in the value for `%USERPROFILE%`? For every character we shorten the longest names, one additional character can appear in this path.

MichaRotstein:
thanks @sharwell , how do I get this variable ?

sharwell:
For me, the path is:

> C:\\Users\\sam\\AppData\\Local\\Microsoft\\VisualStudio\\16.0_80a394a5\\VTC\\45823ca21c203519dc399b38d579e61d\\~PC\\CSharpDiagnostic\\CSharp\\Extensibility\\1033\\CSharpDiagnostic.zip\\ProjectTemplates\\CSharp\\Diagnostic\\Test\\Verifiers\\CSRefactoringVerifier`1+Test.cs

This is 250 characters, so without special system configuration I would expect the template to work for a user with a username up to 13 characters. `Micha Rotstein` is 14 characters. ☹️ Removing one or more characters from the path to this template file should resolve the issue for you.

MichaRotstein:
OK thanks , how do I do that ? (without changing my username)

sharwell:
A proper fix requires modification to the templates themselves, or fixing the IDE to stop producing paths this long in the template cache. In the meantime, the three ways to fix this would be:

1. Generate the project on a machine/account which is shorter, after which time it will work everywhere.
2. Manually create the missing file(s) based on the expected content
3. Delete **CSharpRefactoringVerifier`1.cs** and the VB equivalent, which means you can't test code refactorings but you might not need this anyway

MichaRotstein:
Sorry but I can't see how any of these solutions will work for me, I have never had these kind of issues with any other template.

Is there a way to provide us with a project based on the template without any modifications that builds ?

sharwell:
Here are the templates for the two files that failed to generate:

[...]

PR #749, opened 2021-03-05
Remove the arity suffix from template items

sharwell:
This is an easy way to get a couple more characters allowed in long user names.

Related to #601


genlu:
This also workaround the vssdk issue of backtick in path mentioned here
https://github.com/dotnet/arcade/pull/7056

jmarolf (approved):

PR #759, opened 2021-03-11
Add ProjectState.GeneratedSources API

### LINKED ISSUE #754: It is hard to use test framework for source generators
from https://github.com/dotnet/roslyn-sdk/issues/602#issuecomment-793645429

It is hard to use test framework for source generators. Some `SourceGeneratorTest<TVerifier>` API are not suitable for source generators.

https://github.com/dotnet/roslyn-sdk/blob/50e2a5526c05033b743d8835a259ca4d8851c441/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest%601.cs#L61-L79

## It is hard to use `FixedCode`

It is hard to use `FixedCode` because `FixedState.Sources` needs original codes and generated codes.

## `Verify.NotEmpty` should be removed

`Verify.NotEmpty` should be removed. Source generators can work in a project without any C# code like [CsvGenerator](https://github.com/dotnet/roslyn-sdk/blob/50e2a5526c05033b743d8835a259ca4d8851c441/samples/CSharp/SourceGenerators/SourceGeneratorSamples/CsvGenerator.cs). Even if `TestState.Sources` is empty, test can be valid.

## The generated source code is not always deterministic

The generated source code is not always deterministic when the source generator is using `System.Guid`, `System.Random`,`System.DateTime` etc. 

The output can also change due to the indefinite order of `Dictionary<TKey, TValue>` and `ImmutableDictionary<TKey, TValue>`. 

```csharp
private static StringBuilder CreateMetadataSource(StringBuilder sb, ImmutableDictionary<string, string> metadatas)
{
    sb.AppendLine("using System.Reflection;");
    foreach (var p in metadatas)
    {
        sb.Append("[assembly: AssemblyMetadataAttribute(");
        sb.Append(ToLiteral(p.Key));
        sb.Append(",");
        sb.Append(ToLiteral(p.Value));
        sb.AppendLine(")]");
    }
    return sb;
}
```

### Name of source file

Name of genereted file is especially problematic. 

In Windows, name of generated file is `{AssemblyName}\{ClassName}\{hintName}`, but in Mac/Linux, the name is `{AssemblyName}/{ClassName}/{hintName}`.

sharwell:
* Add `ProjectState.GeneratedSources`. This is the primary property that will eventually allow any test (analyzer, code fix, refactoring, or source generator) to include generated sources.
* Add `TestBehaviors.SkipGeneratedSourcesCheck`, which allows tests to implicitly include source generator outputs (as opposed to verifying that they match the contents of `GeneratedSources`).
* Rewrite source generator testing to no longer verify diagnostics for the intermediate state prior to running source generators.

Closes #754

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
@@ -259,7 +259,7 @@ namespace Microsoft.CodeAnalysis.Testing
                 return;
             }
 
-            if (!expected.Any(x => IsSubjectToExclusion(x, sources)))
+            if (!expected.Any(x => IsSubjectToExclusion(x, analyzers, sources)))
             {
                 return;
             }
@@ -268,7 +268,7 @@ namespace Microsoft.CodeAnalysis.Testing
             // still be reported. We also insert a new line at the beginning so we have to move all diagnostic
             // locations which have a specific position down by one line.
             var expectedResults = expected
-                .Where(x => !IsSubjectToExclusion(x, sources))
+                .Where(x => !IsSubjectToExclusion(x, analyzers, sources))
                 .Select(x => IsInSourceFile(x, sources) ? x.WithLineOffset(1) : x)
                 .ToArray();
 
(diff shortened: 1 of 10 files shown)


jmarolf (approved):

sharwell:
@KathleenDollard @clairernovotny this is a big change to the source generator testing API which makes testing much simpler. It should be in signed builds within the hour. You can see examples of the change impact in [SourceGeneratorValidationTests.cs](https://github.com/dotnet/roslyn-sdk/pull/759/files#diff-7e4b02673944efc6e78ec8c61795253ca6b536a6b52686dfc756a591cca8ba47); let me know if you have any questions about updating.

PR #825, opened 2021-05-19
use nuget packages instead of expecting visual studio to be installed

Youssef1313:
Related to https://github.com/dotnet/samples/pull/4593.

cc @jmarolf @sharwell @BillWagner


sharwell:
> use nuget packages instead of expecting visual studio to be installed

This is already how the samples behaved, so it's not yet clear why a change would be needed.

Youssef1313:
@sharwell Based on what I see on the dotnet/samples PR, this change causes the sample to be built successfully without Roslyn workload being installed. The snippets5000 build on dotnet/samples used to timeout before this change.

But no idea whether the change can cause any undesired behavior or not.

Youssef1313:
Closing this out until the conversation on the other PR is resolved.

Issue #823, opened 2021-05-19
Add API to Microsoft.CodeAnalysis.SourceGenerators.Testing to allow for providing GlobalOptions 


jmarolf:
see example here: https://github.com/jmarolf/generator-start/pull/11

would be nice if the user could provide a set of global options for testing.

CC: @sharwell

sharwell:
This API was added in #734

jmarolf:
The intent here is for people who want to test consumption of [MSBuild Properties](https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md#consume-msbuild-properties-and-metadata). Should probably probably update this to make that clear. Someone could specify this information via editorconfig files but if the intent is to check an msbuild property what do we recommend?

sharwell:
MSBuild properties are injected via a global config, so it would be tested using `AnalyzerConfigFiles` with this format:

```
is_global = true

// properties here
```

PR #827, opened 2021-05-20
Add ability to specify global options in a test

jmarolf:
closes https://github.com/dotnet/roslyn-sdk/issues/823

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest`1.cs
@@ -6,11 +6,9 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
-using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.CodeAnalysis.CodeActions;
-using Microsoft.CodeAnalysis.CodeRefactorings;
+
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Formatting;
 using Microsoft.CodeAnalysis.Simplification;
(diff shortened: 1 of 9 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest`1.cs
```diff
@@ -22,6 +20,11 @@ namespace Microsoft.CodeAnalysis.Testing
     public abstract class SourceGeneratorTest<TVerifier> : AnalyzerTest<TVerifier>
         where TVerifier : IVerifier, new()
     {
+        /// <summary>
+        /// Gets a collection of global options that will be awaiable from Context.AnalyzerConfigOptions.GlobalOptions
+        /// </summary>
+        public List<(string key, string value)> GlobalOptions { get; } = new List<(string, string)>();
```
sharwell, line 26: This is already provided by TestState.AnalyzerConfigFiles.

sharwell (changes requested):

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest`1.cs
jmarolf: what about msbuild files? would our approach then be to have the user supply the props/targets files?

sharwell: No, the testing focuses on the compiler component of a build.

jmarolf: we support adding nuget packages which is not a compiler component of a build. I don't think API is what we want but we don't think providing a way to test this makes sense?

sharwell: I don't understand the request here. We already have the ability to test global options through an existing API. Why would we need to add a new one?

PR #830, opened 2021-05-24
add vb generator samples to the samples solution

jmarolf:


Samples.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28606.18
+# Visual Studio Version 17
```
chsienki, line 3: This change ok?

chsienki (approved):

Samples.sln
jmarolf: oops

PR #896, opened 2021-09-15
Update component debugger to work for VS 2022

chsienki:
Update the component debugger to work in the new launch profiles format introduced in VS2022. See https://github.com/dotnet/project-system/blob/main/docs/repo/property-pages/how-to-add-a-new-launch-profile-kind.md for reference.

Removes the old user control, and writing code which is now handled by the rule file (`ComponentDebuggerLaunchProfile.xaml`). Adds rule exporting and `ProvideBindingPath` so that VS can find the rule and apply it to matching projects.

Tested on VS2022 using an existing `launchsettings.json` and confirmed projects show the new UI, and can add/remove as needed.


chsienki:
@jmarolf @sharwell for review please :)

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Roslyn.ComponentDebugger.csproj
```diff
@@ -8,13 +8,33 @@
...
+    <XamlPropertyRule Include="ComponentDebuggerLaunchProfile.xaml">
+      <CopyToOutputDirectory>Never</CopyToOutputDirectory>
+      <Generator>MSBuild:GenerateRuleSourceFromXaml</Generator>
+      <SubType>Designer</SubType>
+      <DataAccess>None</DataAccess>
+      <RuleInjection>None</RuleInjection>
+    </XamlPropertyRule>
+  </ItemGroup>
+  <Target Name="EmitAssemblyVersion" BeforeTargets="CoreCompile">
+    <WriteLinesToFile File="$(BaseIntermediateOutputPath)\AssemblyVersion.g.cs"
```
chsienki, line 22: 🤢This is unpleasant, but required for the rules. It seems overkill to make it a generator, but I'm considering writing one to handle the whole rule injection which would remove a lot of the ceremony needed here.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/RuleExporter.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Roslyn.ComponentDebugger
+{
+    internal static class RuleExporter
+    {
+        /// <summary>
+        /// Used to export the XAML rule via MEF
+        /// </summary>
+        [ExportPropertyXamlRuleDefinition(
+            xamlResourceAssemblyName: "Roslyn.ComponentDebugger, Version=" + AssemblyVersion.Version + ", Culture=neutral, PublicKeyToken=31bf3856ad364e35",
```
chsienki, line 16: The attribute requires the FQN so we need to embed the version, which comes from the msbuild task.

src/VisualStudio.Roslyn.SDK/Roslyn.SDK/BindingPathProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.VisualStudio.Shell;
+
+namespace Roslyn.SDK
+{
+    /// <summary>
+    /// Ensures that Visual Studio probes the package directory for assemblies at load time.
+    /// </summary>
+    /// <seealso cref="https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.providebindingpathattribute"/>
+    [ProvideBindingPath]
+    internal class BindingPathProvider
```
chsienki, line 14: We need a way to tell VS to look in the package directory to find assemblies for the component debugger rule. We don't have a package for the debugger itself, we just embed it in the SDK package, so this adds the required registry entries for it. Annoyingly you can't put it at the assembly level so we have to have a dummy class to put it on.

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Roslyn.ComponentDebugger.csproj
sharwell: I would prefer a source generator here. Nerdbank.GitVersioning might allow this to be automated without a source generator or project file changes.

@AArnott does NB.GV support setting the AssemblyVersion via the project file instead of the normal approach of automatically determining it from the repo history?

src/VisualStudio.Roslyn.SDK/Roslyn.SDK/BindingPathProvider.cs
sharwell: ❗ Use `ProvideCodeBase` instead of `ProvideBindingPath`

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Roslyn.ComponentDebugger.csproj
AArnott: No. NB.GV will steamroll any `AssemblyVersion` property you set with the discovered version. 
If you wrote build authoring to execute your target in between `GenerateAssemblyVersionInfo` and `GetBuildVersion` then you could rewrite that property. But that strikes me as making your car float instead of building a boat.

Have you considered letting NB.GV do its thing? :wink:

src/VisualStudio.Roslyn.SDK/Roslyn.SDK/BindingPathProvider.cs
chsienki: Ok, I have no idea, was just following the instructions in the doc linked in the description. What's the difference / what does it do?

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Roslyn.ComponentDebugger.csproj
chsienki: The project system is using NB.GV for this exact purpose. I was just trying to avoid adding dependencies for such a small thing, if it supports it, happy to do so if you prefer.

jmarolf:
@drewnoakes can you also take a look here?

jmarolf (approved):
:shipit:

src/VisualStudio.Roslyn.SDK/ComponentDebugger/Roslyn.ComponentDebugger.csproj
chsienki: Updated to use a SG

drewnoakes (approved):
Looks good to me.

src/VisualStudio.Roslyn.SDK/Roslyn.SDK/BindingPathProvider.cs
sharwell: > What's the difference / what does it do?

ProvideBindingPath means every assembly load will look in this extension folder to see if it contains the assembly. ProvideCodeBase restricts the new search location to only apply to one specific assembly (better performance).

chsienki: Gotcha, thanks for the explanation!

PR #902, opened 2021-09-20
Add package reference for System.Net.Http

Cosifne:
A workaround to fix issues: 
https://devdiv.visualstudio.com/DevDiv/_componentGovernance/112744/alert/2980237?typeId=6257816
https://devdiv.visualstudio.com/DevDiv/_componentGovernance/112744/alert/3537864?typeId=6257816

Roslyn-SDK is not using System.Net.Http packages, but because System.Net.Http is involved in NetStandard.Library 1.6.1, and it is restored by default by all NetStandard projects (which includes our projects, and many other libraries)

Explicitly add System.Net.Http 4.3.4 would fix it.

Apparently, the security check tool should not report this, but it is going to block the build starts from 9/30 so we might need this workaround.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Directory.Build.props
+++ b/src/Microsoft.CodeAnalysis.Testing/Directory.Build.props
@@ -88,7 +88,7 @@
   <!-- Public API Analyzer configuration -->
   <ItemGroup>
     <PackageReference Include="Roslyn.Diagnostics.Analyzers" Version="$(RoslynDiagnosticsAnalyzersVersion)" PrivateAssets="all" />
-    <PackageReference Include="System.Net.Http" Version="$(SystemNetHttpVersion)" PrivateAssets="all" />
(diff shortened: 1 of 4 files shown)


src/Microsoft.CodeAnalysis.Testing/Directory.Build.props
```diff
@@ -88,7 +88,7 @@
   <!-- Public API Analyzer configuration -->
   <ItemGroup>
     <PackageReference Include="Roslyn.Diagnostics.Analyzers" Version="$(RoslynDiagnosticsAnalyzersVersion)" PrivateAssets="all" />
-
+    <PackageReference Include="System.Net.Http" Version="$(SystemNetHttpVersion)" PrivateAssets="all" />
```
sharwell, line 91: Can we also include `ExcludeAssets="all"` to make sure the build doesn't reference anything from this library (since the library is not included in the declared dependencies of the package)?

sharwell (approved):

PR #915, opened 2021-10-02

### LINKED ISSUE #914: Extend analyzer testing framework to support DiagnosticSuppressor tests
With the current testing framework it's not possible to write tests for `DiagnosticSuppressor` analyzers.

Suggestion:
Extend `DiagnosticResult` with a property `IsSuppressed` that can be used by the diagnostic verifier to assert the state of the diagnostic.

In the test:
```C#
var results = new[]
{
    DiagnosticResult.CompilerError("CA1822").WithLocation(0).WithIsSuppressed(true)
};

await VerifyCS.VerifyAnalyzerAsync(test, results);
```
In the verifier:
```C#
private class Test : CSharpAnalyzerTest<TAnalyzer, XUnitVerifier>
{
    public Test()
    {
        DiagnosticVerifier = (diagnostic, result, verifier) =>
        {
            var shouldBeSuppressed = result.IsSuppressed.GetValueOrDefault();
            verifier.Equal(shouldBeSuppressed, diagnostic.IsSuppressed, $"Diagnostic {result} => suppression state should be {shouldBeSuppressed}");
        };

```

tom-englert:
Fixes #914

Live sample usage can be found here: https://github.com/tom-englert/Lazy.Fody/blob/master/Lazy.Fody.Analyzer.Tests/SuppressorTests.cs

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/DiagnosticResult.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/DiagnosticResult.cs
@@ -57,7 +57,8 @@ namespace Microsoft.CodeAnalysis.Testing
             DiagnosticOptions options,
             string id,
             LocalizableString? messageFormat,
-            object?[]? messageArguments)
+            object?[]? messageArguments,
+            bool? isSuppressed)
         {
             _spans = spans;
             _suppressMessage = suppressMessage;
@@ -67,6 +68,7 @@ namespace Microsoft.CodeAnalysis.Testing
             Id = id;
             MessageFormat = messageFormat;
             MessageArguments = messageArguments;
+            IsSuppressed = isSuppressed;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


tom-englert:
Does not feel complete yet, since `DiagnosticResult` now offers some property that is never used in the verification by the default implementation.

I have tried to extend the `VerifyDiagnosticResults` method, but `Microsoft.CodeAnalysis.Testing` actually depends on an older version of `Microsoft.CodeAnalysis.Common`, so the referenced version of the `Diagnostic` class does not contain the new `IsSuppressed` property.

Suggested solutions: 
- Bump the version of `Microsoft.CodeAnalysis.Common`. I'm not aware what the strategy is for versioning the dependencies and required backward compatibility, so I would need some advice what version can be used.
- Use reflection to read the `IsSuppressed` property. Not my favorite, but since this only affects testing and not production code, maybe a valid workaround.
- Rename `bool DiagnosticResult.IsSuppressed` to something like `object DiagnosticResult.CustomData`, so it's clear this is only used to pass custom information to your private extension of the analyzer test implementation, and the implementation is responsible for interpreting this data according to the context of the test. This would also make the analyzer test open for other scenarios.

tom-englert:
Implemented scenario 1, by conditionally selecting dependencies and extend the code.

However this seems to have reviled some code flaws that have not been visible before.

tom-englert:
Implemented scenario 2 in #917, by using reflection to get the suppressed property value.

tom-englert:
Superseeded by #917

PR #916, opened 2021-10-04

tom-englert:
This implements scenario 1 from #915  by conditionally selecting dependencies and extend the analyzer validation code.

### DIFF
--- a/Roslyn-SDK.sln
+++ b/Roslyn-SDK.sln
@@ -20,6 +20,10 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Roslyn.SyntaxVisualizer.Extension", "src\VisualStudio.Roslyn.SDK\SyntaxVisualizer\Roslyn.SyntaxVisualizer.Extension\Roslyn.SyntaxVisualizer.Extension.csproj", "{BF3477AA-22C9-4751-A4C0-01EF30728AE9}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Microsoft.CodeAnalysis.Testing", "Microsoft.CodeAnalysis.Testing", "{A3E08CE3-2358-4D36-875B-82C99AC61CD8}"
+	ProjectSection(SolutionItems) = preProject
+		src\Microsoft.CodeAnalysis.Testing\Directory.Build.props = src\Microsoft.CodeAnalysis.Testing\Directory.Build.props
+		src\Microsoft.CodeAnalysis.Testing\Directory.Build.targets = src\Microsoft.CodeAnalysis.Testing\Directory.Build.targets
+	EndProjectSection
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.CodeAnalysis.Analyzer.Testing", "src\Microsoft.CodeAnalysis.Testing\Microsoft.CodeAnalysis.Analyzer.Testing\Microsoft.CodeAnalysis.Analyzer.Testing.csproj", "{36F8BB88-BBDB-4917-BCF4-269E7652A314}"
 EndProject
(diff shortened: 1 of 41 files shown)


src/Microsoft.CodeAnalysis.Testing/Directory.Build.props
```diff
@@ -25,6 +25,23 @@
...
+      <PropertyGroup>
+        <Testing_MicrosoftCodeAnalysisWorkspaces_Version>1.0.1</Testing_MicrosoftCodeAnalysisWorkspaces_Version>
+        <Testing_MicrosoftCodeAnalysisAnalyzers_Version>2.6.1</Testing_MicrosoftCodeAnalysisAnalyzers_Version>
+      </PropertyGroup>
+    </When>
+    <Otherwise>
+      <PropertyGroup>
+        <Testing_MicrosoftCodeAnalysisWorkspaces_Version>3.8.0</Testing_MicrosoftCodeAnalysisWorkspaces_Version>
+        <Testing_MicrosoftCodeAnalysisAnalyzers_Version>3.3.2</Testing_MicrosoftCodeAnalysisAnalyzers_Version>
+        <DefineConstants Condition="'$(Language)' == 'C#'">$(DefineConstants);DIAG_SUPPORTS_SUPPRESSION</DefineConstants>
```
sharwell, line 39: ❗ These can't change. Light-up/reflection must be used to support the scenario in cases where the APIs do not exist.

sharwell (changes requested):

Roslyn-SDK.sln
```diff
@@ -20,6 +20,10 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Roslyn.SyntaxVisualizer.Extension", "src\VisualStudio.Roslyn.SDK\SyntaxVisualizer\Roslyn.SyntaxVisualizer.Extension\Roslyn.SyntaxVisualizer.Extension.csproj", "{BF3477AA-22C9-4751-A4C0-01EF30728AE9}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Microsoft.CodeAnalysis.Testing", "Microsoft.CodeAnalysis.Testing", "{A3E08CE3-2358-4D36-875B-82C99AC61CD8}"
+	ProjectSection(SolutionItems) = preProject
+		src\Microsoft.CodeAnalysis.Testing\Directory.Build.props = src\Microsoft.CodeAnalysis.Testing\Directory.Build.props
+		src\Microsoft.CodeAnalysis.Testing\Directory.Build.targets = src\Microsoft.CodeAnalysis.Testing\Directory.Build.targets
+	EndProjectSection
```
sharwell, line 26: 💡 These don't need to be added to the solution. They are visible in Solution Explorer with this button:

image

tom-englert: It's not about viewing the file, it's about getting FindInFiles with scope solution working as expected. (for guys like me being not so used to all the features and tweaks in an unknown solution)
However I won't insist in adding this...

sharwell: I find it's always a losing battle to keep solution files synchronized with the solution. The only way it's _always_ consistent is to omit them all and point users to the solution filter button. 😄

src/Microsoft.CodeAnalysis.Testing/Directory.Build.props
tom-englert: OK, then I will do another PR using reflection.

sharwell: We have a few places where it's already used. See `CodeActionExtensions` for an example:

https://github.com/dotnet/roslyn-sdk/blob/27fe3d6060150031b532cc7255259c40a37a5fda/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/CodeActionExtensions.cs#L15-L24

sharwell: Also the support for AnalyzerConfigDocuments https://github.com/dotnet/roslyn-sdk/pull/734

tom-englert:
Replaced by #917

PR #917, opened 2021-10-04

### LINKED ISSUE #914: Extend analyzer testing framework to support DiagnosticSuppressor tests

tom-englert:
Fixes #914 
This implements scenario 2 from #915 by reading the IsSuppressed property via reflection

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/DiagnosticExtensions.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/DiagnosticExtensions.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Testing.Extensions
+{
+    internal static class DiagnosticExtensions
+    {
+        private static readonly MethodInfo? IsSuppressedGetMethod = typeof(Diagnostic).GetProperty("IsSuppressed")?.GetMethod;
+
+        public static bool TryGetIsSuppressed(this Diagnostic diagnostic, out bool value)
+        {
+            value = false;
+
+            var rawValue = IsSuppressedGetMethod?.Invoke(diagnostic, null);
+
+            if (rawValue?.GetType() != typeof(bool))
+            {
+                return false;
+            }
+
+            value = (bool)rawValue;
+
+            return true;
+        }
+    }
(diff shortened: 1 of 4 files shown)


Youssef1313:
FYI @roji

tom-englert:
@sharwell any chance to get this into a release?

sharwell:
@tom-englert Can we get some tests added to show how this works?

tom-englert:
Live sample usage can be found here: https://github.com/tom-englert/Lazy.Fody/blob/master/Lazy.Fody.Analyzer.Tests/SuppressorTests.cs

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/DiagnosticExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Reflection;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Testing.Extensions
+{
+    internal static class DiagnosticExtensions
+    {
+        private static readonly MethodInfo? IsSuppressedGetMethod = typeof(Diagnostic).GetProperty("IsSuppressed")?.GetMethod;
+
+        public static bool TryGetIsSuppressed(this Diagnostic diagnostic, out bool value)
```
sharwell, line 16: 💭 It probably isn't necessary to make this `TryGetIsSuppressed`. It could instead follow the pattern from `ProjectExtensions.AnalyzerConfigDocuments`:

1. Cache a `Func<Diagnostic, bool>` instead of `MethodInfo`
2. If the `IsSuppressed` property isn't present, treat its value as `false` instead of missing (diagnostics could not be suppressed prior to diagnostic suppressors)

sharwell (commented):
Overall, the approach seems fine. I was hoping to get some unit tests added to this repository (e.g. similar to how **AnalyzerConfigFilesTests.cs** was added to cover test support for the AnalyzerConfig feature).

sharwell (approved):

PR #922, opened 2021-10-19
Fix long commands breaking component debugger:

chsienki:
- Write command lines out to a response file
- Pass response file to csc instead of raw arguments
- Make component IDisposable and remove file on dispose


sharwell:
@chsienki I had to merge main into this PR to get the build to use the correct image (where integration tests run). Feel free to rebase if you prefer to have it just be one commit.

PR #936, opened 2021-12-13
Update lightup and tests for Diagnostic.IsSuppressed

sharwell:
@tom-englert this look good to you?

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
@@ -412,15 +412,8 @@ namespace Microsoft.CodeAnalysis.Testing
 
                 if (expected.IsSuppressed.HasValue)
                 {
-                    if (actual.diagnostic.TryGetIsSuppressed(out var actualValue))
-                    {
-                        message = FormatVerifierMessage(analyzers, actual.diagnostic, expected, $"Expected diagnostic suppression state to match");
-                        verifier.Equal(expected.IsSuppressed.Value, actualValue, message);
-                    }
-                    else
-                    {
-                        throw new NotSupportedException("DiagnosticSuppressors are not supported on this platform.");
-                    }
+                    message = FormatVerifierMessage(analyzers, actual.diagnostic, expected, $"Expected diagnostic suppression state to match");
+                    verifier.Equal(expected.IsSuppressed.Value, actual.diagnostic.IsSuppressed(), message);
                 }
 
                 DiagnosticVerifier?.Invoke(actual.diagnostic, expected, verifier);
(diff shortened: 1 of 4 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -412,15 +412,8 @@ private void VerifyDiagnosticResults(IEnumerable<(Project project, Diagnostic di
...
-                    if (actual.diagnostic.TryGetIsSuppressed(out var actualValue))
-                    {
-                        message = FormatVerifierMessage(analyzers, actual.diagnostic, expected, $"Expected diagnostic suppression state to match");
-                        verifier.Equal(expected.IsSuppressed.Value, actualValue, message);
-                    }
-                    else
-                    {
-                        throw new NotSupportedException("DiagnosticSuppressors are not supported on this platform.");
-                    }
+                    message = FormatVerifierMessage(analyzers, actual.diagnostic, expected, $"Expected diagnostic suppression state to match");
```
tom-englert, line 415: Definitely improves the code if we do not have to handle this edge-case.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/DiagnosticExtensions.cs
```diff
@@ -3,30 +3,28 @@
...
-using System.Collections.Generic;
 using System.Reflection;
-using System.Text;
 
 namespace Microsoft.CodeAnalysis.Testing.Extensions
 {
     internal static class DiagnosticExtensions
     {
-        private static readonly MethodInfo? IsSuppressedGetMethod = typeof(Diagnostic).GetProperty("IsSuppressed")?.GetMethod;
+        private static readonly Func<Diagnostic, bool> s_isSuppressed;
```
tom-englert, line 12: We could just add the `CreateDelegate` to the field initializer, instead of introducing a static constructor, but I think that's just personal flavor.
```c#
        private static readonly Func<Diagnostic, bool> s_isSuppressed = (Func<Diagnostic, bool>?)typeof(Diagnostic).GetProperty(nameof(IsSuppressed), typeof(bool))
            ?.GetMethod?.CreateDelegate(typeof(Func<Diagnostic, bool>), target: null)
        ?? (_ => false);
```

tom-englert:
👍

jmarolf (approved):

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/DiagnosticExtensions.cs
sharwell: I find the current form a bit more readable. 👍

PR #939, opened 2021-12-14
Log IDE crashes

sharwell:
* Log integration test errors that occur outside of devenv
* Improve integration test stability by updating to include microsoft/vs-extension-testing#91


JoeRobich (commented):
👍

PR #948, opened 2022-01-19
Add extension point for creating a CompilationWithAnalyzers object.

jkoritzinsky:
For the DllImportGenerator unit tests, we'd like to create our own `CompilationWithAnalyzers` object with a `CompilationWithAnalyzersOptions` object to provide callbacks for exception cases. Instead of providing a method to create a `CompilationWithAnalyzersOptions` object, I propose we add a method to create a `CompilationWithAnalyzers` object for two reasons:

1. `CompilationWithAnalyzersOptions` doesn't exist until Microsoft.CodeAnalysis 3.0, but the testing framework supports back to Microsoft.CodeAnalysis 1.x.
2. There's no public API for creating a `CompilationWithAnalyzers` object with both a `CompilationWithAnalyzersOptions` object and a `CancellationToken`.

For the `DllImportGenerator` unit tests, we don't care about the `CancellationToken`, but I don't want to assume that is the case for any other analyzer's testing.

cc: @sharwell for review
cc: @agocke, @dotnet/interop-contrib we can use this API once it's available to help identify any of the intermittent test failures in DllImportGenerator.Unit.Tests we're seeing that might be due to GC holes.


sharwell (approved):

sharwell:
📝 I'm checking to see if the same change is needed in any other place...

PR #970, opened 2022-02-22
Add build step to generate SBOM

JoeRobich:
Test Build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5782883&view=results (microsoft)

SBOM should be in build artifacts.


genlu (approved):
I didn't actually look at the file content though

PR #973, opened 2022-03-14
Provide a way for consumers to not treat . in a markup file as special

CyrusNajmabadi:
I need this as i write more tests for raw strings.  In particular, i tried all sorts of different approaches to deal with things like `$$"""..."""` as well as `$$$$"""..."""` and it just got enormously messy and incredibly hard to even read the test.  I think fundamentally there just has to be a way for a test to say: do not under any circumstances treat `$$` special.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/TestMarkupParser.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/TestMarkupParser.cs
@@ -1,0 +1,406 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics;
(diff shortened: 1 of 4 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/TestMarkupParser.cs
```diff
@@ -0,0 +1,406 @@
...
+
+            var currentIndexInInput = 0;
+            var inputOutputOffset = 0;
+
+            var matches = new List<(int position, string key)>(6);
+            while (true)
+            {
+                matches.Clear();
+
+                if (!_ignorePositionString)
```
CyrusNajmabadi, line 179: this is the only logic add.  everything else is extraction from one type to another.  i tried to just make this so we had one type, with an instance API and a static API.  However, C# doesn't allow the sigs to match (evne if one is instance and one is static).  so i had to pull into another.

CyrusNajmabadi:
I found a horrific hack for thi  on the roslyn side.  Just swap out all `$` signs with a character unlikely to conflict with anything else (like `\uD7FF`).  Then swap those back after running through the parser.  this works :(  you can close htis i fyou want.

PR #975, opened 2022-03-28
Support validating providing fils in nested paths within the Analyzer and CodeFix test suites.

jkoritzinsky:
This enables developers to acurately vaildate that their code-fixes append to files in subfolders of a project (such as Properties/AssemblyInfo.cs) using the same detection mechanism that would work in VS, the Folders property on the various \*Document types.

We're working on implementing a code-fix relating to source-generated interop that either creates a `Properties/AssemblyInfo.cs` file with an attribute or appends to it, and we can't accurately test the "append" case without the testing libraries accurately portraying the `Folders` property.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
@@ -7,6 +7,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Text;
@@ -1194,7 +1195,6 @@ namespace Microsoft.CodeAnalysis.Testing
                     additionalProjectId,
                     solution.GetProject(additionalProjectId).CompilationOptions
                         .WithOutputKind(projectState.OutputKind)
-                        .WithXmlReferenceResolver(xmlReferenceResolver)
                         .WithAssemblyIdentityComparer(referenceAssemblies.AssemblyIdentityComparer));
 
                 solution = solution.WithProjectParseOptions(
@@ -1209,19 +1209,22 @@ namespace Microsoft.CodeAnalysis.Testing
                 foreach (var (newFileName, source) in projectState.Sources)
                 {
                     var documentId = DocumentId.CreateNewId(additionalProjectId, debugName: newFileName);
-                    solution = solution.AddDocument(documentId, newFileName, source, filePath: newFileName);
+                    var (fileName, folders) = GetNameAndFoldersFromPath(newFileName);
+                    solution = solution.AddDocument(documentId, fileName, source, folders: folders, filePath: newFileName);
                 }
 
                 foreach (var (newFileName, source) in projectState.AdditionalFiles)
                 {
                     var documentId = DocumentId.CreateNewId(additionalProjectId, debugName: newFileName);
-                    solution = solution.AddAdditionalDocument(documentId, newFileName, source, filePath: newFileName);
+                    var (fileName, folders) = GetNameAndFoldersFromPath(newFileName);
+                    solution = solution.AddAdditionalDocument(documentId, fileName, source, folders: folders, filePath: newFileName);
                 }
 
                 foreach (var (newFileName, source) in projectState.AnalyzerConfigFiles)
                 {
                     var documentId = DocumentId.CreateNewId(additionalProjectId, debugName: newFileName);
-                    solution = solution.AddAnalyzerConfigDocument(documentId, newFileName, source, filePath: newFileName);
+                    var (fileName, folders) = GetNameAndFoldersFromPath(newFileName);
+                    solution = solution.AddAnalyzerConfigDocument(documentId, fileName, source, folders: folders, filePath: newFileName);
                 }
             }
 
@@ -1230,19 +1233,22 @@ namespace Microsoft.CodeAnalysis.Testing
             foreach (var (newFileName, source) in primaryProject.Sources)
             {
                 var documentId = DocumentId.CreateNewId(projectId, debugName: newFileName);
-                solution = solution.AddDocument(documentId, newFileName, source, filePath: newFileName);
+                var (fileName, folders) = GetNameAndFoldersFromPath(newFileName);
+                solution = solution.AddDocument(documentId, fileName, source, folders: folders, filePath: newFileName);
             }
 
             foreach (var (newFileName, source) in primaryProject.AdditionalFiles)
             {
                 var documentId = DocumentId.CreateNewId(projectId, debugName: newFileName);
-                solution = solution.AddAdditionalDocument(documentId, newFileName, source, filePath: newFileName);
+                var (fileName, folders) = GetNameAndFoldersFromPath(newFileName);
+                solution = solution.AddAdditionalDocument(documentId, fileName, source, folders: folders, filePath: newFileName);
             }
 
             foreach (var (newFileName, source) in primaryProject.AnalyzerConfigFiles)
             {
                 var documentId = DocumentId.CreateNewId(projectId, debugName: newFileName);
-                solution = solution.AddAnalyzerConfigDocument(documentId, newFileName, source, filePath: newFileName);
+                var (fileName, folders) = GetNameAndFoldersFromPath(newFileName);
+                solution = solution.AddAnalyzerConfigDocument(documentId, fileName, source, folders: folders, filePath: newFileName);
             }
 
             solution = AddProjectReferences(solution, projectId, primaryProject.AdditionalProjectReferences.Select(name => projectIdMap[name]));
@@ -1262,6 +1268,13 @@ namespace Microsoft.CodeAnalysis.Testing
             static Solution AddProjectReferences(Solution solution, ProjectId sourceProject, IEnumerable<ProjectId> targetProjects)
             {
                 return solution.AddProjectReferences(sourceProject, targetProjects.Select(id => new ProjectReference(id)));
+            }
+
+            static (string fileName, IEnumerable<string> folders) GetNameAndFoldersFromPath(string path)
+            {
+                string fileName = Path.GetFileName(path);
+                string[] folders = Path.GetDirectoryName(path)!.Split(Path.DirectorySeparatorChar, StringSplitOptions.RemoveEmptyEntries);
+                return (fileName, folders);
             }
         }
 
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Runtime.ExceptionServices;
 using System.Threading;
@@ -527,7 +528,7 @@ namespace Microsoft.CodeAnalysis.Testing
                 verifier.EqualOrDiff(newState.Sources[i].content.ToString(), actual.ToString(), $"content of '{newState.Sources[i].filename}' did not match. Diff shown with expected as baseline:");
                 verifier.Equal(newState.Sources[i].content.Encoding, actual.Encoding, $"encoding of '{newState.Sources[i].filename}' was expected to be '{newState.Sources[i].content.Encoding?.WebName}' but was '{actual.Encoding?.WebName}'");
                 verifier.Equal(newState.Sources[i].content.ChecksumAlgorithm, actual.ChecksumAlgorithm, $"checksum algorithm of '{newState.Sources[i].filename}' was expected to be '{newState.Sources[i].content.ChecksumAlgorithm}' but was '{actual.ChecksumAlgorithm}'");
-                verifier.Equal(newState.Sources[i].filename, updatedDocuments[i].Name, $"file name was expected to be '{newState.Sources[i].filename}' but was '{updatedDocuments[i].Name}'");
+                verifier.True(FileNamesEqual(newState.Sources[i].filename, updatedDocuments[i].Folders, updatedDocuments[i].Name), $"file name was expected to be '{newState.Sources[i].filename}' but was '{updatedDocuments[i].Name}'");
             }
 
             var updatedAdditionalDocuments = project.AdditionalDocuments.ToArray();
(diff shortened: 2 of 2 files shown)


sharwell (changes requested):
Marking Request Changes as a workaround for GitHub not supporting required reviewers.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1209,19 +1209,22 @@ protected virtual async Task<Project> CreateProjectImplAsync(EvaluatedProjectSta
                 foreach (var (newFileName, source) in projectState.Sources)
                 {
                     var documentId = DocumentId.CreateNewId(additionalProjectId, debugName: newFileName);
-                    solution = solution.AddDocument(documentId, newFileName, source, filePath: newFileName);
+                    var (fileName, folders) = GetNameAndFoldersFromPath(newFileName);
+                    solution = solution.AddDocument(documentId, fileName, source, folders: folders, filePath: newFileName);
```
sharwell, line 1213: ❓ I've long suspected our construction of the document info differed slightly from what occurs in practice. Can you provide debugger proof that the update here matches real-world analyzer cases for all four impacted properties?

* `debugName` (part of `DocumentId`)
* `name`
* `folders`
* `filePath`

I believe a correct is required here but want to be certain we fix this for good. 😄

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1194,7 +1195,6 @@ protected virtual async Task<Project> CreateProjectImplAsync(EvaluatedProjectSta
                     additionalProjectId,
                     solution.GetProject(additionalProjectId).CompilationOptions
                         .WithOutputKind(projectState.OutputKind)
-                        .WithXmlReferenceResolver(xmlReferenceResolver)
```
sharwell, line 1197: ❓ Why is this getting removed?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1263,6 +1269,13 @@ static Solution AddProjectReferences(Solution solution, ProjectId sourceProject,
             {
                 return solution.AddProjectReferences(sourceProject, targetProjects.Select(id => new ProjectReference(id)));
             }
+
+            static (string fileName, IEnumerable<string> folders) GetNameAndFoldersFromPath(string path)
+            {
+                var fileName = Path.GetFileName(path);
+                var folders = Path.GetDirectoryName(path)!.Split("/");
```
sharwell, line 1276: ❓ Is it possible for the path to show up with `\` separators? We use `/` by default in the test framework because it simplifies cross-platform testing, but it seems possible some users are passing in `\` here for individual tests (e.g. by using `Path.Combine`).

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
```diff
@@ -553,7 +554,19 @@ private async Task VerifyProjectAsync(ProjectState newState, Project project, IV
...
                 verifier.Equal(newState.AnalyzerConfigFiles[i].content.Encoding, actual.Encoding, $"encoding of '{newState.AnalyzerConfigFiles[i].filename}' was expected to be '{newState.AnalyzerConfigFiles[i].content.Encoding?.WebName}' but was '{actual.Encoding?.WebName}'");
                 verifier.Equal(newState.AnalyzerConfigFiles[i].content.ChecksumAlgorithm, actual.ChecksumAlgorithm, $"checksum algorithm of '{newState.AnalyzerConfigFiles[i].filename}' was expected to be '{newState.AnalyzerConfigFiles[i].content.ChecksumAlgorithm}' but was '{actual.ChecksumAlgorithm}'");
-                verifier.Equal(newState.AnalyzerConfigFiles[i].filename, updatedAnalyzerConfigDocuments[i].Name, $"file name was expected to be '{newState.AnalyzerConfigFiles[i].filename}' but was '{updatedAnalyzerConfigDocuments[i].Name}'");
+                verifier.True(FileNamesEqual(newState.AnalyzerConfigFiles[i].filename, updatedAnalyzerConfigDocuments[i].Folders, updatedAnalyzerConfigDocuments[i].Name), $"file name was expected to be '{newState.AnalyzerConfigFiles[i].filename}' but was '{GetFileName(updatedAnalyzerConfigDocuments[i].Folders, updatedAnalyzerConfigDocuments[i].Name)}'");
```
sharwell, line 557: 💡 For all seven cases (three in this file and four elsewhere), let's extract the result of `GetFileName` to a local variable, inline the call to `FileNamesEqual`, and return to using `verifier.Equal`.

sharwell (changes requested):

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
jkoritzinsky: If users are overriding `DefaultFilePathPrefix` or passing in custom documents, it's possible for `\` to get into the string.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
jkoritzinsky: I've validated this already for simple cases, but I'll circle back to see how various different edge cases are represented in VS or Omnisharp (specifically files out-of-tree in VS and absolute paths on both) and update the implementation accordingly.

jkoritzinsky: Here's the info I've found for VS:

| File location | `debugName` | `name` | `folders` | `filePath` |
|--------------|-----------------|--------|----------|-----------|
| In subdirectory of project included with relative path | absolute path | File name with extension | folder path relative to the project root | absolute path |
| Absolute path to file | Absolute path | file name with extension | empty list | file name and extension | absolute path |
|  Relative path to file outside of project tree | Absolute path | file name with extension | empty list | file name and extension | absolute path |
|  File outside of project tree with `Link` metadata | absolute path | file name with extension | folders from `Link` metadata | original file name and extension (ignores a new name provided in `Link` metadata) | absolute path |

I think the only accurate way to portray this is to extend the `SourceFileCollection` API to also track folder paths and debug name, but since that's an API-breaking change, I don't want to do that without your agreement. Do you think that's the best approach, or should we try to shoehorn in a best-effort solution with the existing API surface?

sharwell: The project root can be determined as:

* If `ProjectState.DefaultPrefix` is not an absolute path, there is no project root.
* If `ProjectState.DefaultPrefix` ends with `/` or `\`, this is the project root.
* Otherwise, `Path.GetDirectoryName(ProjectState.DefaultPrefix)` is the project root.

jkoritzinsky: That sounds good to me. I'll update the implementation to do that.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1194,7 +1195,6 @@ protected virtual async Task<Project> CreateProjectImplAsync(EvaluatedProjectSta
                     additionalProjectId,
                     solution.GetProject(additionalProjectId).CompilationOptions
                         .WithOutputKind(projectState.OutputKind)
```
jkoritzinsky, line 1197:
```suggestion
                        .WithOutputKind(projectState.OutputKind)
                        .WithXmlReferenceResolver(xmlReferenceResolver)
```

jkoritzinsky:
@sharwell this is ready for another round of review. I've updated the design to match your suggestions.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1265,6 +1272,46 @@ static Solution AddProjectReferences(Solution solution, ProjectId sourceProject,
...
+        {
+            if (Path.IsPathRooted(path))
+            {
+                // If the user provides a rooted path as the file name, just use that as-is.
+                return (path, NoFoldersArray);
+            }
+
+            // Normalize to platform path separators for simplicity later on
+            var normalizedPath = path.Replace('\\', Path.DirectorySeparatorChar);
+            var normalizedDefaultPathPrefix = DefaultFilePath.Replace('\\', Path.DirectorySeparatorChar);
```
sharwell, line 1285: ⚠️ `DefaultFilePath` only applies to the top-most `ProjectState`. The `AdditionalProjects` collection contains other projects, each with its own default file path. This method should accept the default file path as an argument, and the caller updated to use the one for the project it is currently processing.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1265,6 +1272,46 @@ static Solution AddProjectReferences(Solution solution, ProjectId sourceProject,
...
+            if (!Path.IsPathRooted(normalizedDefaultPathPrefix))
+            {
+                // If our default path isn't rooted, then we assume that we don't have any rooted paths
+                // and just use the file name
+                return (Path.GetFileName(normalizedPath), NoFoldersArray);
```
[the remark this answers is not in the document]
jkoritzinsky: `Array.Empty` doesn’t exist in net452 (which this library still targets).

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1265,6 +1272,47 @@ static Solution AddProjectReferences(Solution solution, ProjectId sourceProject,
...
+            var normalizedDefaultPathPrefix = projectPathPrefix.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
+
+            if (!Path.IsPathRooted(normalizedDefaultPathPrefix))
+            {
+                // If our default path isn't rooted, then we assume that we don't have any rooted paths
+                // and just use the file name
+                return (Path.GetFileName(normalizedPath), NoFoldersArray);
+            }
+
+            var projectRootPath = Path.GetFileName(normalizedDefaultPathPrefix) == string.Empty
```
sharwell, line 1288: 💡 A comment on this would be good.

| Default path | Project root path |
| --- | --- |
| `/0/Temp` | `/0/` |
| `/0/` | `/0/` |

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1265,6 +1272,47 @@ static Solution AddProjectReferences(Solution solution, ProjectId sourceProject,
...
+                ? normalizedDefaultPathPrefix
+                : Path.GetDirectoryName(normalizedDefaultPathPrefix)!;
```
sharwell, line 1290:
```suggestion
                : (Path.GetDirectoryName(normalizedDefaultPathPrefix) + Path.DirectorySeparatorChar);
```

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1265,6 +1272,47 @@ static Solution AddProjectReferences(Solution solution, ProjectId sourceProject,
...
+                }
+
+                // Otherwise, to match VS behavior we will report no folders and only the file name.
+                return (Path.GetFileName(normalizedPath), NoFoldersArray);
+            }
+
+            var subpath = normalizedPath.Substring(projectRootPath.Length);
+
+            var fileName = Path.GetFileName(subpath);
+            var folders = Path.GetDirectoryName(subpath)!.Split(Path.DirectorySeparatorChar);
```
sharwell, line 1310: ❗ For the specific case where the file is in the project folder, we can't use `Split` because it returns `new[] { "" }` instead of an empty array.

```suggestion
            if (Path.GetDirectoryName(subpath) == string.Empty)
            {
                return (fileName, folders: new string[0]);
            }

            var folders = Path.GetDirectoryName(subpath)!.Split(Path.DirectorySeparatorChar);
```

sharwell (changes requested):

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest`1.cs
```diff
@@ -91,13 +93,29 @@ protected async Task<ImmutableArray<Diagnostic>> VerifySourceGeneratorAsync(Solu
...
-                    verifier.Equal(expectedSources[i].filename, updatedDocuments[i].Name, $"file name was expected to be '{expectedSources[i].filename}' but was '{updatedDocuments[i].Name}'");
+
+                    // Source-generated sources are implicitly in a subtree, so they have a different folders calculation.
+                    var (fileName, folders) = i < numOriginalSources
+                        ? GetNameAndFoldersFromPath(DefaultFilePathPrefix, expectedSources[i].filename)
+                        : GetNameAndFoldersFromSourceGeneratedFilePath(expectedSources[i].filename);
+                    verifier.Equal(fileName, updatedDocuments[i].Name, $"file name was expected to be '{fileName}' but was '{updatedDocuments[i].Name}'");
+                    verifier.SequenceEqual(folders, updatedDocuments[i].Folders, message: $"folders was expected to be '{string.Join(DirectorySeparatorString, folders)}' but was '{string.Join(DirectorySeparatorString, updatedDocuments[i].Folders)}'");
```
sharwell, line 102:
```suggestion
                    verifier.SequenceEqual(folders, updatedDocuments[i].Folders, message: $"folders was expected to be '{string.Join("/", folders)}' but was '{string.Join("/", updatedDocuments[i].Folders)}'");
```

sharwell (commented):
A few tweaks but overall looking good

sharwell (approved):
Looks good. I'll submit a separate fix for integration tests.

jkoritzinsky:
Sounds good. I don't have merge permissions in this repo, so merge it when you think it's ready.

jkoritzinsky:
Now that CI is green, can we merge this in? One of our features for sounds generated interop requires this for accurate testing.

Issue #974, opened 2022-03-21
ArgumentNullException from NuGet calls in Microsoft.CodeAnalysis.Testing.ReferenceAssemblies.ResolveAsync


jkoritzinsky:
We recently hit an `ArgumentNullException` that stemmed from us using the `ReferenceAssemblies` API in dotnet/runtime. This occurred in https://github.com/dotnet/runtime/pull/66952 on macOS x64 ([Helix Log here](https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-66952-merge-bbd9bbac458c42438b/LibraryImportGenerator.Unit.Tests/1/console.d33f14e2.log?sv=2019-07-07&se=2022-04-10T19%3A36%3A01Z&sr=c&sp=rl&sig=vd2BMsqzgNVoC6OWpbIb4XbkmqAOww9ZmTzn9SbzShU%3D)).

This looks like it may have been a one-off failure, but I wanted to report it in case it's reproducible as even one-off-style failures can cause significant issues keeping dotnet/runtime green.

jkoritzinsky:
Another hit in https://dev.azure.com/dnceng/public/_build/results?buildId=1674834&view=ms.vss-test-web.build-test-results-tab&runId=45943994&resultId=151060&paneView=debug, so this isn't a one-off.

cc: @sharwell

jkoritzinsky:
We're hitting this pretty often in dotnet/runtime, so it's definitely not a one-off failure.

danmoseley:
This is now happening 10x a day in dotnet/runtime   @jaredpar @sharwell could someone please take a look?


```kusto
TestResults
| join kind=inner WorkItems on WorkItemId
| join kind=inner Jobs on JobId
| where Finished >= now(-2d)
| where Result == "Fail"
| where Message startswith "System.ArgumentNullException : Value cannot be null. (Parameter 'package"
| project Type, Method,
  Pipeline = tostring(parse_json(Properties).DefinitionName),//WorkItemFriendlyName,
  Pipeline_Configuration = tostring(parse_json(Properties).configuration),
  OS = QueueName,
  Arch = tostring(parse_json(Properties).architecture),
 Finished,
 Build
```

|Type|Method|Pipeline|Pipeline_Configuration|OS|Arch|Finished|Build|
|---|---|---|---|---|---|---|---|
[...]

PR #977, opened 2022-04-04
Handle case of null packageStream

danmoseley:
Fix https://github.com/dotnet/roslyn-sdk/issues/974

TBH, I do not know whether this is the right fix (or even a fix). 

I am following this pattern:
https://github.com/nuget/nuget.client/blob/0ff5f67097522299620f34bd13f92d3d58e41f93/src/NuGet.Core/NuGet.PackageManagement/Projects/FolderNuGetProject.cs#L166-L188

review without whitespace changes
https://github.com/dotnet/roslyn-sdk/pull/977/files?diff=unified&w=1

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
@@ -355,16 +355,33 @@ namespace Microsoft.CodeAnalysis.Testing
                             continue;
                         }
 
-                        await PackageExtractor.ExtractPackageAsync(
+                        if (downloadResult.Status == DownloadResourceResultStatus.AvailableWithoutStream)
+                        {
+                            await PackageExtractor.ExtractPackageAsync(
 #if !NET452 && !NETSTANDARD1_5
 #pragma warning disable SA1114 // Parameter list should follow declaration
-                            downloadResult.PackageSource,
+                                downloadResult.PackageSource,
 #pragma warning restore SA1114 // Parameter list should follow declaration
 #endif
-                            downloadResult.PackageStream,
-                            localPathResolver,
-                            packageExtractionContext,
-                            cancellationToken);
+                                downloadResult.PackageReader,
+                                localPathResolver,
+                                packageExtractionContext,
+                                cancellationToken);
+                        }
+                        else
+                        {
+                            Debug.Assert(downloadResult.PackageStream != null, "PackageStream should not be null if download result status != DownloadResourceResultStatus.AvailableWithoutStream");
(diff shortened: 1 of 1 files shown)


jkoritzinsky (approved):
Given the NuGet lInk you shared earlier, LGTM. cc: @sharwell

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -355,16 +355,33 @@ private async Task<ImmutableArray<MetadataReference>> ResolveCoreAsync(string la
...
+                        else
+                        {
+                            Debug.Assert(downloadResult.PackageStream != null, "PackageStream should not be null if download result status != DownloadResourceResultStatus.AvailableWithoutStream");
+                            await PackageExtractor.ExtractPackageAsync(
+#if !NET452 && !NETSTANDARD1_5
+#pragma warning disable SA1114 // Parameter list should follow declaration
+                                downloadResult.PackageSource,
+#pragma warning restore SA1114 // Parameter list should follow declaration
+#endif
+                                downloadResult.PackageStream,
```
sharwell, line 380: In the reference link, this path passes _both_ the PackageReader and the PackageStream. Should we be doing that?

danmoseley: That's the 3rd overload, and it would require stream not be null too:
https://github.com/nuget/nuget.client/blob/0ff5f67097522299620f34bd13f92d3d58e41f93/src/NuGet.Core/NuGet.Packaging/PackageExtractor.cs#L152

Again though I really am just pattern matching here. We could commit this on the assumption that clearly it would only affect a path that's failing already. Or, we could consult someone on the Nuget team, what do you prefer?

danmoseley:
@sharwell the test failures seem to be pre-existing, I see them in https://dev.azure.com/dnceng/public/_build/results?buildId=1695926&view=ms.vss-test-web.build-test-results-tab&runId=46277452&resultId=100002&paneView=debug for https://github.com/dotnet/roslyn-sdk/pull/975/checks?check_run_id=5795657798 ?

sharwell:
Yes, I'm fixing the test failures in #976

danmoseley:
Rerunning now tests fixed

danmoseley:
@sharwell  now failing thus 

error : Failed to verify the root directory of local source 'https://pkgs.dev.azure.com/dnceng/public/&#95;packaging/dotnet-tools/nuget/v3/index.json'.
The given path's format is not supported.

danmoseley:
@sharwell can you please review?

sharwell (approved):

PR #981, opened 2022-04-18
feat: Added ConfigFile to simplify AnalyzerConfigFile creation.

HavenDV:
I recently got acquainted with your package and for me creating a configuration file was problematic until I found an example. I think this can simplify the entry threshold because it adds a specific Add method that will be used intuitively.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,52 @@
+# roslyn-sdk Agent Context
+
+This file gives Codex and other agents enough repository context to work safely and to support later portfolio analysis.
+
+<!-- BEGIN GENERATED PORTFOLIO CONTEXT -->
+
+## Generated Repository Context
+
+This section was generated from the GitHub repository inventory and local checkout to support future Codex work and portfolio analysis.
+
+### Repository Metadata
+- Remote: https://github.com/HavenDV/roslyn-sdk
+- Visibility: public
+- Type: fork; active
+- Primary language: C#
+- Topics: None detected
+- Last pushed: 2022-04-20T23:16:24Z
+- Local path: /Users/havendv/GitHub/HavenDV/roslyn-sdk
+- Local note: standard checkout
+- Classification: Fork/reference workspace
+
+### Working Summary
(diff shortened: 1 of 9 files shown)


sharwell:
I'm a bit hesitant to add this here in the current form. Note that you can achieve the same outcome by defining the `Add` method as an extension method in your own project.

Here are issues I'd like to see addressed in a general API:

1. Allow adding `/.editorconfig` or `/.globalconfig` without specifying the file name directly.
2. Avoid having the new `Add` method apply to `ProjectState.Sources`, `ProjectState.AdditionalFiles`, or `ProjectState.GeneratedSources`.
3. Avoid the need to type `new Dictionary<string, Dictionary<string, string>>` or `new Dictionary<string, string>`, similar to how we avoid the need to type `new ProjectState(projectName, language, $"/{projectName}/Test", extension)` when adding items to `SolutionState.AdditionalProjects`.

sharwell (changes requested):
See notes in review

HavenDV:
Thank you. Tried to take into account your comments.
The only downside I don't like about the new implementation is the duplication of the section key here if you need to add multiple values to the same section.
```cs
var actual = new EditorConfigFile
{
    ["root"] = "true",
    ["*", "key"] = "{|Literal:value|}",
}.ToString();
```

sharwell:
> The only downside I don't like about the new implementation is the duplication of the section key here if you need to add multiple values to the same section.

I did notice this. I believe you could resolve this by making `root` be a constructor parameter:

```csharp
var actual = new EditorConfigFile(root: true)
{
  ["*"] =
  {
    "key = {|Literal:value|}",
  },
}.ToString();
```

or:

```csharp
var actual = new EditorConfigFile(root: true)
{
  ["*"] =
  {
    { "key", "{|Literal:value|}" },
  },
}.ToString();
```

HavenDV:
Perhaps I do not understand something, but how to specify global values other than root, which can be anything?

Will the values for the `*` section be available via `AnalyzerConfigOptionsProvider.GlobalOptions`?

dibarbet:

PR #988, opened 2022-05-18
Use weighted match to allow expected generator outputs in any order

sharwell:


tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing.UnitTests/SourceGeneratorValidationTests.cs
```diff
@@ -36,6 +36,42 @@ public async Task AddSimpleFile()
             }.RunAsync();
         }
 
+        [Fact]
+        public async Task MultipleFilesAllowEitherOrder()
```
AArnott, line 40: This test (or another) seems like it would be more interesting if the generated files didn't have exactly the same content, and empty besides.

AArnott (approved):

AArnott:
I won't pretend to have thoroughly understood the algorithm deeply, but the change looks good. Especially the recent optimizations.

JoeRobich (approved):

PR #1001, opened 2022-07-22
Propogate AdditionalReferences in project state.

jkoritzinsky:
This enables tests that need to pass additional metadata references to additional projects that are passed to the AnalyzerTest testing harness (as well as all derived types)

Fixes #1000

PR #1002, opened 2022-07-28
Fix duplicate PackageReference in Restore

jkoritzinsky:
Fix duplicate PackageReference in restore that is breaking the build in other PRs and local builds.

Fixes the build in #1001.


jkoritzinsky:
cc: @jmarolf @sharwell

jmarolf (approved):

jkoritzinsky:
@jmarolf the same integration test failed on the release job on both attempts. Is this a known flaky test?

PR #1003, opened 2022-08-01
Provide an extensibility point in AnalyzerTest for sorting diagnostics.

jkoritzinsky:
The sort here is non-deterministic for any analyzers whose diagnostics can differ only on the message format and properties.

Allow the sorting algorithm to be changed by the user to include either or both of these elements if they emit diagnostics that differ only on these values from a concurrent analyzer (the .NET Interop team owns an analyzer like this, which prompted this work). This will ensure that everyone can ensure that their diagnostic and code-fix tests (which rely on the ordering to determine the order for the iterative code-fix tests and the fix-all tests) are deterministic.

cc: @sharwell


mavasani (approved):

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1317,14 +1317,14 @@ protected virtual Workspace CreateWorkspaceImpl()
         /// <param name="diagnostics">A collection of <see cref="Diagnostic"/>s to be sorted.</param>
         /// <returns>A collection containing the input <paramref name="diagnostics"/>, sorted by
         /// <see cref="Diagnostic.Location"/> and <see cref="Diagnostic.Id"/>.</returns>
-        private static (Project project, Diagnostic diagnostic)[] SortDistinctDiagnostics(IEnumerable<(Project project, Diagnostic diagnostic)> diagnostics)
+        protected virtual ImmutableArray<(Project project, Diagnostic diagnostic)> SortDistinctDiagnostics(IEnumerable<(Project project, Diagnostic diagnostic)> diagnostics)
```
mavasani, line 1320: @jkoritzinsky You need to update the only callsite of this method to not invoke `ToImmutableArray`: https://github.com/dotnet/roslyn-sdk/blob/1df5d394195ffcf6aab5b824ae420197ed1f0acc/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest%601.cs#L932-L933

Feel free to merge after making that change.

PR #1026, opened 2022-10-10
run integration tests in separate job

jmarolf:
Integration tests are regularly taking longer than the timeout. Want to split them out

### DIFF
--- a/.vsts-pr.yaml
+++ b/.vsts-pr.yaml
@@ -34,16 +34,8 @@ jobs:
         _args: -test
         _configuration: Debug
         _solution: Roslyn-SDK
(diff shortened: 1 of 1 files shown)


JoeRobich (approved):

JoeRobich:
Looking at previous passing integration test runs and they seem to complete both `-test` and `-integrationTest` in <30min. Have you run these integration tests locally?

jmarolf:
> Looking at previous passing integration test runs and they seem to complete both `-test` and `-integrationTest` in <30min. Have you run these integration tests locally?

yep, it appears integration tests just hang indefinitely now. I am going to disable them while I investigate what has changed.

PR #1039, opened 2022-11-10
Add support in AnalyzerTest to include source generators as part of the test run

### LINKED ISSUE #1038: Add support in AnalyzerTest to include source generators as part of the test run
Currently, analyzers that require a source generator to run so that the actual source 
compiles successfully, cannot be run because only solution transforms are available.

By adding an `AnalyzerTest.CompilationTransforms` list (similar to `SolutionTransforms`) 
property, test authors could easily apply additional source generators to the compilation 
before the rest of the test evaluation continues, for example:

```csharp
        var test = new AnalyzerTest
        {
            TestCode = "...",
            CompilationTransforms =
            {
                (compilation, project) =>
                {
                    CSharpGeneratorDriver
                        .Create(new StaticGenerator())
                        .RunGeneratorsAndUpdateCompilation(compilation, out var staticPass, out var _);

                    CSharpGeneratorDriver
                        .Create(new IncrementalGenerator())
                        .RunGeneratorsAndUpdateCompilation(staticPass, out var incrementalPass, out var _);

                    return incrementalPass;
                }
            }
        };
```

With this in place, the analyzer can verify code that uses generated code in order to properly 
compile.

kzu:
Implements `CompilationTransforms` so that test authors can provide additional transforms after the basic compilation has been created.

Closes #1038

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,11 +1,11 @@
 {
   "sdk": {
-    "version": "7.0.100-preview.6.22352.1",
+    "version": "7.0.100",
     "allowPrerelease": true,
-    "rollForward": "major"
+    "rollForward": "latestFeature"
   },
   "tools": {
-    "dotnet": "7.0.100-preview.6.22352.1",
+    "dotnet": "7.0.100",
     "runtimes": {
       "dotnet": [
         "3.1.0"
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
@@ -150,6 +150,12 @@ namespace Microsoft.CodeAnalysis.Testing
         public Action<Diagnostic, DiagnosticResult, IVerifier>? DiagnosticVerifier { get; set; }
 
         /// <summary>
+        /// Gets a collection of transformation functions to apply to a <see cref="Compilation"/> during diagnostic or code
+        /// fix test setup.
+        /// </summary>
+        public List<Func<Compilation, Project, Compilation>> CompilationTransforms { get; } = new List<Func<Compilation, Project, Compilation>>();
+
+        /// <summary>
         /// Gets a collection of transformation functions to apply to <see cref="Workspace.Options"/> during diagnostic
         /// or code fix test setup.
         /// </summary>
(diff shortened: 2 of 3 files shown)


jmarolf:
@kzu thanks!, I'll take a look

sharwell (changes requested):
I believe it would be better to do the following:

1. Update the implementation derived from `AnalyzerTest` in your project to have a `SourceGenerators` property
2. Override `GetProjectCompilationAsync` to apply the source generators before returning, similar to how the [source generator](https://github.com/dotnet/roslyn-sdk/blob/ba5f8a6bf7b477b44a43ea943d3e831d6301afb7/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest%601.cs#L50-L54) test framework does it

I'll look into updating the test framework to support this approach directly.

sharwell:
@kzu The framework was designed with inheritance in mind, which is why the project templates default to creating the derived types. Note that you don't need to create a large number of derived types; it's generally sufficient to create one derived type that is used to correctly configure the test framework for all the common test cases in your library.

kzu:
BTW, the SourceGenerators testing library does not support incremental source generators, so this is even more pressing that might otherwise seem....

sharwell:
@kzu Incremental source generators are supported by updating the local derived verifier instance, like this:

https://github.com/microsoft/vs-extension-testing/tree/main/src/Microsoft.VisualStudio.Extensibility.Testing.SourceGenerator.UnitTests/Verifiers

I'm going to go ahead and close this PR since we're interested in the feature, but will need a significant rewrite to deliver it. The tracking feature request issues should still be open.

kzu:
There's a constraint in the test class that only includes ISourceGenerator, not IIncrementalGenerator: https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing/CSharpSourceGeneratorTest%602.cs#L13

Unless you're saying (indirectly) that the  test-derived classes shouldn't be used at all for incremental generator tests?

sharwell:
The example verifier I linked is constrained to `IIncrementalGenerator`.

sharwell:
I believe that is already answered above. It would be better if the library supported source generators in the analyzer layer, but so far we haven't had the time to deliver it. Whenever we do deliver this feature, it will follow the common patterns in this library that encourage the use of inheritance to customize the behavior within the context of a downstream test project.

PR #1042, opened 2022-12-09
Add new api that allows you to specify a package source

jmarolf:
This makes it much simpler to construct reference assemblies for un-released packages


sharwell:
Has this been tested? Other uses of this feature appear to be for setting local sources (e.g. a folder on disk).

bogdanbeldiemorningstar:
h

PR #1048, opened 2023-01-18
Update Microsoft.CodeAnalysis.Analyzers to 3.3.4

Youssef1313:


Directory.Build.props
```diff
@@ -5,6 +5,7 @@
   <PropertyGroup>
     <Features>strict</Features>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
```
sharwell: Context on this?

Youssef1313: @sharwell

1. There is a new diagnostic `RS1036` which gets fired in analyzer/generator projects that don't set `EnforceExtendedAnalyzerRules`.
2. If `EnforceExtendedAnalyzerRules` is set to true, we get banned API diagnostic for specific APIs defined in https://github.com/dotnet/roslyn-analyzers/blob/4512de66bb6e21c548ab0d5a83242b70969ba576/src/Microsoft.CodeAnalysis.Analyzers/Core/AnalyzerBannedSymbols.txt, and if it's false, we don't get any diagnostics.

sharwell: Seems like we should set the property to true in analyzer projects (presumably in the project files of individual projects), and not set it at all in other projects.

Youssef1313: @sharwell It must always have a value. Otherwise we get RS1036 (I don't think we want to suppress it?)

We could default to `false` and set to `true` in analyzer projects. That's what Manish suggested in dotnet/roslyn actually.

Youssef1313: My thought was "why not default to true if it gets green"

sharwell: Sure, that would work too.

sharwell (approved):

PR #1051, opened 2023-01-27
Update .NET 7 reference assemblies to latest stable

Youssef1313:


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -987,7 +987,7 @@ public static class Net
                         "net7.0",
                         new PackageIdentity(
                             "Microsoft.NETCore.App.Ref",
-                            "7.0.0-preview.7.22375.6"),
+                            "7.0.2"),
```
sharwell, line 990: These would typically lock at the initial stable release.

```suggestion
                            "7.0.0"),
```

Since they are reference assemblies, it's unlikely a patch would change public surface area.

sharwell (approved):

PR #1052, opened 2023-01-27
Update environment variable overrides

sharwell:
* Use new extension testing library feature to clear .NET-related CI environment variables
* Update environment variable clearing to include more variables

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -11,7 +11,9 @@
     <UsingToolMicrosoftNetCompilers>true</UsingToolMicrosoftNetCompilers>
     <UsingToolSymbolUploader>true</UsingToolSymbolUploader>
     <UsingToolNuGetRepack>true</UsingToolNuGetRepack>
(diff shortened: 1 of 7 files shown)


.vsts-pr.yaml
```diff
@@ -61,10 +63,19 @@ jobs:
         publishLocation: Container
       continueOnError: true
       condition: not(succeeded())
+    - task: PublishBuildArtifacts@1
+      displayName: Publish test logs
+      inputs:
+        PathtoPublish: '$(Build.SourcesDirectory)\artifacts\TestResults\$(_configuration)'
+        FileCopyOptions: /xd cloudtest_*
```
sharwell, line 70: ☹️ still haven't figured out how to make this work. Two copies of heap dumps are getting uploaded on failure

ryzngard: I think xcopy options only work if `publishLocation: FilePath` is set instead of `container`

ryzngard (approved):

PR #1058, opened 2023-02-09
Update NuGet references to address security issues

sharwell:


jmarolf (approved):

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -89,8 +89,8 @@ public static ReferenceAssemblies Default
         {
             get
             {
-#if NETSTANDARD1_5
-                return NetStandard.NetStandard15;
+#if NETSTANDARD1_6
+                return NetStandard.NetStandard16;
```
sharwell, line 93: 📝 I'm not thrilled by this, but since the final runtime assembly is selected by the executable project, it seems unlikely that this could will ever be hit. The user would need to be targeting .NET Core 1.x in order for this binary to be selected.

PR #1075, opened 2023-03-25
Fix #1074

### LINKED ISSUE #1074: AnalyzerTesting: Make the generated syntax trees of source generators available
The current approach to test source generators by providing the expected generated code upfront is not very convenient, since the generated sources can be very large, making the test unreadable.

It would be great to support another scenario, where the analyzer test is run with `TestBehaviors.SkipGeneratedSourcesCheck`, and then the generated sources can be fetched and verified with custom logic, e.g. using https://github.com/VerifyTests/ like in this sample https://github.com/Fody/PropertyChanged/blob/dee9d81bab0202dc2e5d985910072ad6e6c94fa4/PropertyChanged.Fody.Analyzer.Tests/CodeGeneratorTest.cs#L55-L71

tom-englert:
Fix #1074: AnalyzerTesting: Make the generated syntax trees of source generators available


sharwell (changes requested):
The generated trees are already available. I use them in microsoft/vs-extension-testing for record/replay testing:
https://github.com/microsoft/vs-extension-testing/blob/b78ac2a63a471cc0a56b45af0a1d7b2473288202/src/Microsoft.VisualStudio.Extensibility.Testing.SourceGenerator.UnitTests/Verifiers/CSharpSourceGeneratorVerifier%601%2BTest.cs#L76-L80

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -168,6 +168,11 @@ public string TestCode
         /// </summary>
         public List<Func<Solution, ProjectId, Solution>> SolutionTransforms { get; } = new List<Func<Solution, ProjectId, Solution>>();
 
+        /// <summary>
+        /// Gets the syntax trees generated by the source generators during the test run.
+        /// </summary>
+        public ImmutableArray<SyntaxTree> GeneratedTrees { get; private set; } = ImmutableArray<SyntaxTree>.Empty;
```
sharwell, line 174: 📝 I have quite a few concerns about this. However, since no changes are required to implement the original feature, it might be easiest to just close this PR.

tom-englert: See my comments in #1074. It should not be necessary to rewrite half of the test framework again and again, just to get hold of this simple thing.

tom-englert: BTW, what are your concerns here?

sharwell, line 174: ❗ Source generators can run more than once during a test. It's not clear if/when/how this property would be valid during the life of a test. I would expect the following conceptual changes:

1. The state of `AnalyzerTest<T>` should not change during the life of a call to `RunAsync`.
2. The `GeneratedTrees` should not appear as a property on `AnalyzerTest<T>`, since that would make initializing the test instance more confusing (the output property would be showing up as an input property).
3. The manner in which generated trees are provided accounts for the fact that generators may run multiple times during a test.

tom-englert: about 1:
it already does:
private readonly ConcurrentBag<Workspace> _workspaces = new ConcurrentBag<Workspace>()

about 2:
It won't, because the setter is private

about 3:
The sample you provided overwrites `GetProjectCompilationAsync()` to get access to the generated sources - but isn't that called also many times, so the problem is exactly the same?

tom-englert: Maybe wrapping the call to GenerateSources into a new virtual method could be a good solution?
It would simplify extracting the sources a lot

tom-englert:
Supposed solution works with acceptable effort

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: > about 1:
> it already does:
> private readonly ConcurrentBag _workspaces = new ConcurrentBag()

This is true, but also specifically designed to account for an unknown number of workspace creations. This was a bit clumsy in its own addition, and hopefully it could be avoided for future cases.

> about 2: It won't, because the setter is private

`TestState` doesn't have a setter at all. The syntax used to initialize `TestState` is also valid for properties that do not have a public setter.

> about 3: The sample you provided overwrites `GetProjectCompilationAsync()` to get access to the generated sources - but isn't that called also many times, so the problem is exactly the same?

It is, but the path that accesses the generated sources is only executed during specific record cases.

> Maybe wrapping the call to GenerateSources into a new virtual method could be a good solution?
It would simplify extracting the sources a lot

`ApplySourceGeneratorAsync` might be the method to expose. The part I don't like is it's a difficult method to override.

PR #1077, opened 2023-03-25
Document architectural constraints, so they are visible to the public

### LINKED ISSUE #1076: Enforce StyleCop rules that will break CI-Build
It's very annoying that StyleCop rules fail the CI-Build, but are only warnings in VisualStudio:

image

tom-englert:
Fix #1076: StyleCop rules that will fail the CI-Build should generate errors already while editing

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -112,3 +112,9 @@ dotnet_style_qualification_for_field = false:none
 dotnet_style_qualification_for_property = false:none
 dotnet_style_qualification_for_method = false:none
 dotnet_style_qualification_for_event = false:none
+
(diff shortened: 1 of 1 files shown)


sharwell (changes requested):
No diagnostic should be set to a severity greater than `warning` except for compiler errors where it is impossible to produce a correct binary. Artificially raising the severity of diagnostics to `error` interferes with the developer inner loop for things like red/green development while running tests.

CI sets the Treat Warnings as Errors flag to ensure that any diagnostics that slip through are caught before merge.

sharwell (approved):

PR #1082, opened 2023-04-01
Fix #1081: Fix nullable annotations

### LINKED ISSUE #1081: Nullability annotations are wrong
Both `TestCode `and `FixedCode` properties accept null, but are marked as not-null

https://github.com/dotnet/roslyn-sdk/blob/d4e8061f34846fe992b2bcc933ab342c5447be71/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest%601.cs#L96-L105

tom-englert:
Fixes #1081


Youssef1313 (commented):
What is the scenario you wanted to do `TestCode = someNullableValue`? I can't think of any reasonable scenario.

sharwell (changes requested):
I believe these properties are already annotated correctly. I provided additional context in #1081.

PR #1094, opened 2023-05-03
Add an experimental API for testing incremental steps.

jkoritzinsky:
Add a .NET 6.0 test target that references Roslyn 4.4 so we have a test suite that references Roslyn 4.X.

Introduce a new API to enable developers to define post-initial-run solution transforms to test incrementality.

Add a Humanizer reference at the same version as Roslyn's dependency to improve the error message wording of the new validation section (can be removed if we so desire).

> **Note**
>
> Notes from @sharwell on specific tests we believe would be valuable. These tests could be provided by the API in this pull request, or through some other means.
>
> * Ability to verify that a change which did not impact outputs only executed _specific_ incremental steps. Since it's easy (per @jkoritzinsky) to unintentionally break equality comparisons for model objects, there is a likelihood of regression in the incremental evaluation due to equivalent items returning not-equal.
> * Ability to verify that a change which only impacted a subset of outputs only executed _specific_ incremental steps that serve as inputs to other outputs which did not change.
> * Ability to test that a source generator did not include `Compilation` as part of its incremental state. In other words, after running source generators on a compilation and still holding the `GeneratorDriver` instance, the `Compilation` passed to it should be eligible for garbage collection.


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -336,6 +352,39 @@ protected async Task<ImmutableArray<Diagnostic>> VerifySourceGeneratorAsync(Solu
...
+                {
+                    secondRunSolution = incrementalTransform(secondRunSolution, initialProject.Id);
+                }
+
+                var secondRunProject = secondRunSolution.GetProject(initialProject.Id)!;
+
+                updatedDriver = updatedDriver
+                        .ReplaceAdditionalTexts(secondRunProject.AnalyzerOptions.AdditionalFiles)
+                        .WithUpdatedParseOptions(secondRunProject.ParseOptions)
+                        .WithUpdatedAnalyzerConfigOptions(secondRunProject.AnalyzerOptions.AnalyzerConfigOptionsProvider());
```
sharwell, line 370: 📝 I don't like how one of these transforms allows for any solution transform, but the generator driver transform only handles a few specific cases. It would be better if `IncrementalGeneratorTransforms` was a list of specific transforms that can be applied to the first compilation and first generator driver together.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -1298,7 +1402,26 @@ where parameterTypes.SequenceEqual(new[] { iincrementalGeneratorType })
...
+            }
+            else
+            {
+                var constructor = generatorDriverOptionsType.GetConstructor(
+                    new[]
+                    {
+                        generatorDriverOptionsType.GetTypeInfo().Assembly.GetType("Microsoft.CodeAnalysis.IncrementalGeneratorOutputKind")!,
+                        typeof(bool),
+                    });
+                var options = constructor!.Invoke(new object[] { 0, true });
```
sharwell, line 1421: This would work with just `Activator.CreateInstance` right?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: Ah, it seems like some of the generator properties are not part of the `Compilation` at all, so they are injected into the driver separately from the call to `RunGenerators`. Weird.

jkoritzinsky: I'll see if I can do a `ProjectState` object here like we do with the `CodeFixTest` code instead of doing solution transforms. Alternatively, we could give the users the `GeneratorDriver` and the `Compilation` and allow them to transform them themselves.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
jkoritzinsky: We can't use `Activator.CreateInstance` here as that doesn't support implicit conversions of integer values to enum types, and we don't have the enum type available. We could use `Convert.ChangeType` with the `0` to the enum type mentioned above so we could use `Activator.CreateInstance`, but I don't know if that's any cleaner than the alternative.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -125,6 +126,8 @@ public string TestCode
 
         public SolutionState TestState { get; }
 
+        public SolutionState IncrementalChangeTestState { get; }
```
sharwell, line 129: 💭 Would it make sense to move this into the source generator testing package? It'd lose the ability to test the incrementality of a source generator and test a code fix _in the same test operation_, but I'm not sure that's a significant limitation.

jkoritzinsky: I'll try to move it around and see what I can do.

jkoritzinsky: I was able to move it down. Let me know what you think.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -281,20 +282,24 @@ protected async Task<ImmutableArray<Diagnostic>> VerifySourceGeneratorAsync(Solu
                 return ImmutableArray<Diagnostic>.Empty;
             }
 
-            return await VerifySourceGeneratorAsync(Language, sourceGenerators, testState, verifier.PushContext("Source generator application"), cancellationToken);
```
sharwell, line 284: 📝 Looks like I forgot a `ConfigureAwait(false)` here.

sharwell: 💭 Actually it looks like I forgot more than one. Just leave it for now and I'll review the situation later. Most of the time in Roslyn test code we omit them, but I need to determine the best plan for this library.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing/CSharpSourceGeneratorTest`2.cs
```diff
@@ -27,5 +28,19 @@ protected override CompilationOptions CreateCompilationOptions()
 
         protected override ParseOptions CreateParseOptions()
             => new CSharpParseOptions(DefaultLanguageVersion, DocumentationMode.Diagnose);
+
+        public IncrementalGeneratorExpectedState IncrementalGeneratorState
```
sharwell, line 32: 💡 I think we can remove this in favor of changing the type of `IncrementalGeneratorStepStates` to a new type derived from `Dictionary<Type, IncrementalGeneratorExpectedState>` that provides a new indexer that creates the value on demand. In practice in means you'd have to type this, but it seems acceptable:

```
IncrementalGeneratorStepStates =
{
    [typeof(TSourceGenerator)] =
    {
        ExpectedStepStates =
        {
            // ...
        },
    },
},
```

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest`1.cs
```diff
@@ -4,14 +4,139 @@
...
+
+            // If any solution transforms for incremental generator testing have been defined,
+            // then test incremental generators.
+            // We'll also only go down here if incremental generators exist.
+            if (incrementalTestState is not null && typeof(CompilationOptions).Assembly.GetType("Microsoft.CodeAnalysis.IIncrementalGenerator") is not null)
+            {
+                var secondRunProject = await CreateProjectAsync(new EvaluatedProjectState(incrementalTestState, ReferenceAssemblies), incrementalTestState.AdditionalProjects.Values.Select(additionalProject => new EvaluatedProjectState(additionalProject, ReferenceAssemblies)).ToImmutableArray(), cancellationToken);
+
+                driver = driver
+                        .ReplaceAdditionalTexts(secondRunProject.AnalyzerOptions.AdditionalFiles)
```
sharwell, line 72: ❔ Will the driver recognize cases where a file did not change? For example, consider the case of two additional files. If you change the content of just one of them, will it treat the other one as unchanged?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest`1.cs
```diff
@@ -4,14 +4,139 @@
...
+        /// </summary>
+        public Dictionary<Type, IncrementalGeneratorExpectedState> IncrementalGeneratorStepStates { get; } = new Dictionary<Type, IncrementalGeneratorExpectedState>();
+
+        protected override async Task RunImplAsync(CancellationToken cancellationToken)
+        {
+            if (!TestState.GeneratedSources.Any())
+            {
+                // Verify the test state has at least one source, which may or may not be generated
+                Verify.NotEmpty($"{nameof(TestState)}.{nameof(SolutionState.Sources)}", TestState.Sources);
+            }
```
Youssef1313, line 38: Isn't it okay to have empty sources but non-empty generated sources?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.SourceGenerators.Testing/SourceGeneratorTest`1.cs
jkoritzinsky: Yes I believe so.

Issue #1045, opened 2023-01-05
CodeFixProvider in default template crashes for top-level statements


Youssef1313:
I think the analyzer should skip top-level type.

PR #1100, opened 2023-06-11
Do not report diagnostic on implicit top-level `Program` type in default template

DoctorKrolic:
Fixes: https://github.com/dotnet/roslyn-sdk/issues/1045

### DIFF
--- a/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Analyzer/DiagnosticAnalyzer.cs
+++ b/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Analyzer/DiagnosticAnalyzer.cs
@@ -41,6 +41,13 @@ namespace $saferootprojectname$
             // TODO: Replace the following code with your own analysis, generating Diagnostic objects for any issues you find
             var namedTypeSymbol = (INamedTypeSymbol)context.Symbol;
 
+            // Check whether the first syntax reference is a type declaration syntax.
+            // This guards against trying to report diagnostic for implicit top-level `Program` type, which cannot be renamed
+            if (!(namedTypeSymbol.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() is BaseTypeDeclarationSyntax))
+            {
+                return;
+            }
+
             // Find just those named type symbols with names containing lowercase letters.
             if (namedTypeSymbol.Name.ToCharArray().Any(char.IsLower))
             {


src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Analyzer/DiagnosticAnalyzer.cs
```diff
@@ -41,6 +41,13 @@ private static void AnalyzeSymbol(SymbolAnalysisContext context)
             // TODO: Replace the following code with your own analysis, generating Diagnostic objects for any issues you find
             var namedTypeSymbol = (INamedTypeSymbol)context.Symbol;
 
+            // Check whether the first syntax reference is a type declaration syntax.
+            // This guards against trying to report diagnostic for implicit top-level `Program` type, which cannot be renamed
+            if (!(namedTypeSymbol.DeclaringSyntaxReferences.FirstOrDefault()?.GetSyntax() is BaseTypeDeclarationSyntax))
```
sharwell, line 46: 💭 it seems like there would be an easier way to do this, but so far I haven't found one

Youssef1313: I even suspect this logic is not correct.

The first declared syntax code be another partial `Program` in which case it still can't be renamed.

The approach I would take is to get compilation entry point in compilation start action, and if its name is `TopLevelStatementsEntryPointMethodName`, store its containing type in a variable. Then in AnalyzeSymbol, I would do a symbol comparison.

It might be too much for a starter template though, but still good to demonstrate how to use compilation start actions actually.

@sharwell thoughts?

DoctorKrolic: Isn't the first syntax reference always the one pointing to top-level statement and other references point to partial definitions? Your approach is 100% correct, but I think it is too much for a default template. Just for reference, I've regularly contributed to roslyn for over a year and I still cannot say that I 100% understand how stateful analyzers work. Bombarding a newcomer with so much information in a template isn't a good idea

Youssef1313: > Isn't the first syntax reference always the one pointing to top-level statement and other references point to partial definitions?

I don't know, but even if true, that would be an implementation detail that shouldn't be relied upon. I always take DeclaringSyntaxReferences ordering as a "no guarantee".

PR #1102, opened 2023-06-20
Bump version to address CVE-2023-29337

RussKie:
See https://dev.azure.com/dnceng/internal/_componentGovernance/dotnet-aspnetcore/alert/7916743?typeId=15414781


sharwell:
We fixed this in #1103

PR #1103, opened 2023-06-22
Update default NuGet reference to 6.3.3

sharwell:


dibarbet (approved):

RussKie:
What'd be the new version with the fix?

sharwell:
> What'd be the new version with the fix?

1.1.2-beta1.23322.1

Issue #1108, opened 2023-07-21
Net70Android is missing


Youssef1313:
In https://github.com/dotnet/roslyn-sdk/pull/1041, it looks like Net70Android was missed.

PR #1109, opened 2023-07-21
Add missing Net70Android

Youssef1313:
Fixes https://github.com/dotnet/roslyn-sdk/issues/1108

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
@@ -995,11 +995,11 @@ namespace Microsoft.CodeAnalysis.Testing
                         ImmutableArray.Create(
                             new PackageIdentity("Microsoft.macOS.Ref", "12.3.2372"))));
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -995,6 +995,12 @@ public static class Net
                         ImmutableArray.Create(
                             new PackageIdentity("Microsoft.macOS.Ref", "12.3.2372"))));
 
+            private static readonly Lazy<ReferenceAssemblies> _lazyNet70Android =
+                new Lazy<ReferenceAssemblies>(() =>
+                    Net70.AddPackages(
+                        ImmutableArray.Create(
+                            new PackageIdentity("Microsoft.Android.Ref.33", "33.0.68"))));
```
sharwell, line 1002: 💭 Is there an easy way to know if this is the correct version?

Youssef1313: @sharwell I just used the latest stable that shows as supporting net7.0 on NuGet.org.
Not sure if you have any concerns, or *why* a version could be wrong?

sharwell (approved):

PR #1139, opened 2024-01-04
Support span and minimum severity in CreateFixAllContext

sharwell:


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
```diff
@@ -241,28 +241,34 @@ protected virtual CodeFixContext CreateCodeFixContext(Document document, TextSpa
         /// Creates a new <see cref="FixAllContext"/>.
         /// </summary>
         /// <param name="document">Document within which fix all occurrences was triggered, or null when applying fix all to a diagnostic with no source location.</param>
+        /// <param name="diagnosticSpan">Span for the diagnostic for which fix all occurrences was triggered.</param>
         /// <param name="project">Project within which fix all occurrences was triggered.</param>
         /// <param name="codeFixProvider">Underlying <see cref="CodeFixes.CodeFixProvider"/> which triggered this fix all.</param>
         /// <param name="scope"><see cref="FixAllScope"/> to fix all occurrences.</param>
         /// <param name="codeActionEquivalenceKey">The <see cref="CodeAction.EquivalenceKey"/> value expected of a <see cref="CodeAction"/> participating in this fix all.</param>
         /// <param name="diagnosticIds">Diagnostic Ids to fix.</param>
+        /// <param name="minimumSeverity">The minimum severity of diagnostics to fix in this operation.</param>
```
mavasani, line 250: Why minimum severity and not the exact severity to fix?

mavasani (approved):

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
elachlan: @mavasani Would the idea be that the fix all only applies to similar severity analyzer results? So that only errors are fixed and not warnings or messages?

I guess the other way is to ignore "silent" and still "fix all".

mavasani: > Would the idea be that the fix all only applies to similar severity analyzer results? So that only errors are fixed and not warnings or messages?

I just wanted to know the rationale behind choosing minimum severity over just the results whose severity matches the severity of the trigger diagnostic for FixAll. I personally would have imagined FixAll to fix all diagnostics with the ID regardless of the severity or just those that match the severity of trigger diagnostic. I don't have a big pushback though, I think such a situation is going to be quite rare anyways.

sharwell: We need the ability to batch fix warnings/errors (which break a build) without making changes to code where diagnostics were reduced to hidden/suggestion for the specific purpose of avoiding code churn.

PR #1142, opened 2024-01-24
Avoid unnecessary M.VisualBasic and S.ValueTuple dependencies

ViktorHofer:
Found in https://github.com/dotnet/runtime/pull/96795#discussion_r1464797169

These packages shouldn't be referenced on modern TFMs as the types are already provided inbox by the framework.

---

FWIW, I'm not sure why roslyn-sdk targets out-of-support TFMs like net452 or net46: https://learn.microsoft.com/de-de/lifecycle/products/microsoft-net-framework. That goes against what Tactics agreed on in November. See https://github.com/dotnet/designs-microsoft/pull/74 for more details.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
@@ -49,10 +49,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="1.0.1" />
-    <PackageReference Include="Microsoft.VisualBasic" Version="10.0.1" />
+    <PackageReference Include="Microsoft.VisualBasic" Version="10.0.1" Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'" />
     <PackageReference Include="Microsoft.VisualStudio.Composition" Version="16.1.8" />
     <PackageReference Include="Nullable" Version="1.3.0" PrivateAssets="all" />
-    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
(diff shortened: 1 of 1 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Microsoft.CodeAnalysis.Analyzer.Testing.csproj
```diff
@@ -49,10 +49,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="1.0.1" />
-    <PackageReference Include="Microsoft.VisualBasic" Version="10.0.1" />
+    <PackageReference Include="Microsoft.VisualBasic" Version="10.0.1" Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'" />
     <PackageReference Include="Microsoft.VisualStudio.Composition" Version="16.1.8" />
     <PackageReference Include="Nullable" Version="1.3.0" PrivateAssets="all" />
```
akoeplinger, line 54: this package is probably also not necessary on netcore

ViktorHofer: No idea, I don't really care about that one as it doesn't get promoted to a package dependency.

akoeplinger: ah yeah, and it seems the package also doesn't include the nullability sources on net5.0+

akoeplinger (approved):

sharwell:
> FWIW, I'm not sure why roslyn-sdk targets out-of-support TFMs like net452 or net46

It's a testing package that supports testing Roslyn 1.x analyzers in Visual Studio 2015.

sharwell (approved):

jaredpar:
> It's a testing package that supports testing Roslyn 1.x analyzers in Visual Studio 2015.

Tactics though decided that we're dropping support for older TFMs in our core packages going forward.

PR #1144, opened 2024-01-31
Remove xunit specific packages

### LINKED ISSUE #1099: xUnit.net support broken with xUnit.net 2.5
In upgrading the [xUnit.net analyzers project](https://github.com/xunit/xunit.analyzers) to use a prerelease 2.5.0 build, I've determined that `Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.XUnit` (and/or some of its dependencies) are broken when my code uses 2.5.

Any verifier which fails breaks because of a breaking binary dependency on several of the xUnit.net exception classes. (The assertion library, including all the exceptions, has been overhauled for v2 2.5 and v3.) The source of the problem is that you're taking a binary dependency on `xunit.assert`, which is locking consumers into your version choice when breaking changes are made to anything you consume.

This behavior isn't appropriate, in my view. Your dependency in essence locks users down to only being able to use "compatible" versions of xUnit.net. This is true for any binary dependency you take outside of the core framework: if you ship a hard dependency on a binary that someone else may want to use in their tests, then you are potentially locking them down to only being able to use a version that's compatible with your choice. I ended up removing _all_ binary dependencies from xUnit.net v2 for exactly this reason; for us, the problematic library ended up being JSON.NET, and it broke JSON.NET when they wanted to test themselves with us. Even if we're one of the few or only consumers of this library, we're guaranteed to want to "move on" to newer versions of xUnit.net. 😄 

Here is a quick example showing the problem. Given the follow code which should clearly fail:

image

This is the (expected) result using xUnit.net 2.4.2:

image

And this is the (incorrect) result using xUnit.net 2.5.0-pre.22:

image

I can think of at least two suggestions for potential paths forward.

### 1. Don't use `xunit.assert` at all.

xUnit.net considers all exceptions to be failing tests. Don't consume the xUnit.net assertion library; instead, create your own exceptions with your own messages and throw those.

### 2. Use `xunit.assert.source` (or Git sub-module) instead.

If you wish to consume the library but not ship a binary dependency on it, you can import the source version of our library and use that instead. The [project](https://github.com/xunit/assert.xunit) is designed to be consumed in both of these fashions.

By default, importing the source (via NuGet or sub-module) will bring in all the code as public, but you can override that with `XUNIT_VISIBILITY_INTERNAL`. You can use [pre-processor directives](https://github.com/xunit/assert.xunit/#annotations) to influence which features are enabled or disabled. By leaving everything as internal, you will be able to use the existing assertions without republishing them into your API.

sharwell:
Closes #1099


chsienki (approved):

src/Microsoft.CodeAnalysis.Testing/README.md
```diff
@@ -46,15 +46,15 @@ To reference prerelease packages, add a **NuGet.Config** file to your solution d
...
+  * Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing
 * Visual Basic
-  * Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing.XUnit
-  * Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing.XUnit
-  * Microsoft.CodeAnalysis.VisualBasic.CodeRefactoring.Testing.XUnit
-  * Microsoft.CodeAnalysis.VisualBasic.SourceGenerators.Testing.XUnit
+  * Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing
+  * Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing
+  * Microsoft.CodeAnalysis.VisualBasic.CodeRefactoring.Testing
+  * Microsoft.CodeAnalysis.VisualBasic.SourceGenerators.Testing
```
jasonmalinowski, line 57: This very much worries me here that we're making a source breaking change more or less for folks who consumed this. It seems a much smaller change would just be to remove the verifier implementation that is causing trouble here, and leave the packages and core scaffolding in place. This way:

1. People who have to migrate their packages forward won't have to change a bunch of package references/usings.
2. If we still have xunit specific value we do want to provide (say, xunit does offer a way to have richer exceptions), we have a natural process for restoring that without asking everybody to change their package references/usings _again_.

sharwell:
For now, superseded by #1145. In the future, if we remove these packages, we would remove all of the framework-specific packages.

sharwell:
I updated this pull request to remove all framework specific packages.

src/Microsoft.CodeAnalysis.Testing/README.md
sharwell: We published a public 1.2 release with these types marked obsolete, finally giving us a path forward to remove them. 🎉

PR #1149, opened 2024-03-02
Add ReferenceAssemblies.Net.Net90

mpidash:
I want to use an API (https://github.com/dotnet/runtime/issues/60393) introduced in .NET 9 for an analyzer that I am writing (https://github.com/dotnet/roslyn-analyzers/pull/6967).

If preview 1 is too early to add it here, I could add it temporarily in roslyn-analyzer (and remove it when Net90 is added later).


buyaa-n:
@sharwell is this blocked with something? Could we merge?

PR #1162, opened 2024-06-13
Obsolete framework-specific packages

sharwell:
* Mark types obsolete
* Update documentation and document migration process
* Update default templates

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/README.md
+++ b/src/Microsoft.CodeAnalysis.Testing/README.md
@@ -17,52 +17,48 @@ To reference prerelease packages, add a **NuGet.Config** file to your solution d
 </configuration>
 ```
 
-### MSTest
+### Packages
 
 * C#
-  * Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.MSTest
-  * Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.MSTest
-  * Microsoft.CodeAnalysis.CSharp.CodeRefactoring.Testing.MSTest
-  * Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing.MSTest
+  * Microsoft.CodeAnalysis.CSharp.Analyzer.Testing
+  * Microsoft.CodeAnalysis.CSharp.CodeFix.Testing
+  * Microsoft.CodeAnalysis.CSharp.CodeRefactoring.Testing
+  * Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing
 * Visual Basic
-  * Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing.MSTest
-  * Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing.MSTest
-  * Microsoft.CodeAnalysis.VisualBasic.CodeRefactoring.Testing.MSTest
-  * Microsoft.CodeAnalysis.VisualBasic.SourceGenerators.Testing.MSTest
+  * Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing
+  * Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing
+  * Microsoft.CodeAnalysis.VisualBasic.CodeRefactoring.Testing
+  * Microsoft.CodeAnalysis.VisualBasic.SourceGenerators.Testing
 
-### NUnit
+### Obsolete Packages
 
-* C#
-  * Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.NUnit
-  * Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.NUnit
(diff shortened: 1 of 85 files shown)


src/Microsoft.CodeAnalysis.Testing/README.md
```diff
@@ -17,52 +17,48 @@ To reference prerelease packages, add a **NuGet.Config** file to your solution d
...
+    Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.MSTest &rarr; Microsoft.CodeAnalysis.CSharp.Analyzer.Testing
+
+* Remove the **.MSTest** suffix from namespaces. For example:
+
+    ```patch
+    -using Microsoft.CodeAnalysis.CSharp.Testing.MSTest;
+    +using Microsoft.CodeAnalysis.CSharp.Testing;
+    ```
+
+* Use `CSharpAnalyzerVerifier<TAnalyzer, DefaultVerifier>` instead of `AnalyzerVerifier<TAnalyzer>`. A similar change
```
333fred, line 51: Could we consider defining a `CSharpAnalyzerVerify<TAnalyzer> : CSharpAnalyzerVerifier<TAnalyzer, DefaultVerifier>` here, to make migration easier?

src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/VisualBasic/Diagnostic/Test/UnitTestProject.vbproj
```diff
@@ -9,12 +9,12 @@
...
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeRefactoring.Testing.MSTest" Version="1.1.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing.MSTest" Version="1.1.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing.MSTest" Version="1.1.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeRefactoring.Testing.MSTest" Version="1.1.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing" Version="1.1.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing" Version="1.1.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeRefactoring.Testing" Version="1.1.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing" Version="1.1.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing" Version="1.1.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeRefactoring.Testing" Version="1.1.0" />
```
333fred, line 17: Do we know what package versions these will be? Can we update them?

sharwell: It's a bit of a catch-22 because updating them now will cause the integration tests to fail

sharwell: If this is important, it's highly likely that we will have the ability to make this change after the new packages are published and before the next time this repo gets inserted for a VS release.

src/Microsoft.CodeAnalysis.Testing/README.md
sharwell: I considered creating an `AnalyzerVerifier<TAnalyzer>` which would bind as soon as the namespace changed, but worried that having two types with the same name would be an even worse situation. Note that the obsolete type is not typically used in a large number of cases, because the default project template creates a set of verifier types derived from the obsolete type directly in the test project, so changing the base type in one location would automatically update all tests using it.

src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/VisualBasic/Diagnostic/Test/UnitTestProject.vbproj
333fred: > If this is important, it's highly likely that we will have the ability to make this change after the new packages are published and before the next time this repo gets inserted for a VS release.

That works for me.

src/Microsoft.CodeAnalysis.Testing/README.md
333fred: Makes sense. Maybe we consider doing that in a future version.

333fred (approved):

PR #1166, opened 2024-06-21
Fix compliation errors in Testing README sample custom verifier

MattKotsenas:
Copy / pasting the sample custom verifier from the bottom of the Testing README results in 9 errors and 2 warnings:

| Severity | Code | Description |
|----------|-------|--------------|
| Error | CS0117 | 'CSharpAnalyzerVerifier<TAnalyzer>.Test' does not contain a definition for 'TestCode' |
| Error | CS1061 | 'CSharpAnalyzerVerifier<TAnalyzer>.Test' does not contain a definition for 'ExpectedDiagnostics' and no accessible extension method 'ExpectedDiagnostics' accepting a first argument of type 'CSharpAnalyzerVerifier<TAnalyzer>.Test' could be found (are you missing a using directive or an assembly reference?) |
| Error | CS1061 | 'CSharpAnalyzerVerifier<TAnalyzer>.Test' does not contain a definition for 'ExpectedDiagnostics' and no accessible extension method 'ExpectedDiagnostics' accepting a first argument of type 'CSharpAnalyzerVerifier<TAnalyzer>.Test' could be found (are you missing a using directive or an assembly reference?) |
| Error | CS1061 | 'CSharpAnalyzerVerifier<TAnalyzer>.Test' does not contain a definition for 'RunAsync' and no accessible extension method 'RunAsync' accepting a first argument of type 'CSharpAnalyzerVerifier<TAnalyzer>.Test' could be found (are you missing a using directive or an assembly reference?) | 
| Error | CS1061 | 'CSharpAnalyzerVerifier<TAnalyzer>.Test' does not contain a definition for 'RunAsync' and no accessible extension method 'RunAsync' accepting a first argument of type 'CSharpAnalyzerVerifier<TAnalyzer>.Test' could be found (are you missing a using directive or an assembly reference?) |
| Error | CS0117 | 'CSharpAnalyzerVerifier<TAnalyzer>.Test' does not contain a definition for 'TestCode' |
| Warning | CS8625 | Cannot convert null literal to non-nullable reference type. |
| Warning | CS8625 | Cannot convert null literal to non-nullable reference type. |
| Error | CS0080 | Constraints are not allowed on non-generic declarations |
| Error | CS1520 | Method must have a return type |
| Error | CS0305 | Using the generic type 'CSharpCodeFixVerifier<TAnalyzer, TCodeFix>' requires 2 type arguments |

Which can be addressed with these minor edits:

1. Remove the default `null` value for `Diagnostic`, given that the delegated method is also non-nullable
2. Fix the location of the generic params on the nested `CSharpCodeFixVerifier.Test` type
3. Remove generic constraints on the inner type `CSharpCodeFixVerifier.Test` (because they belong on the outer type)
4. Fix constructor name on `CSharpCodeFixVerifier.Test`


src/Microsoft.CodeAnalysis.Testing/README.md
```diff
@@ -202,7 +202,7 @@ use of "basic use cases" for test scenarios that would otherwise be considered a
 public static class CSharpAnalyzerVerifier<TAnalyzer>
     where TAnalyzer : DiagnosticAnalyzer, new()
 {
-    public static DiagnosticResult Diagnostic(string diagnosticId = null)
+    public static DiagnosticResult Diagnostic(string diagnosticId)
         => CSharpAnalyzerVerifier<TAnalyzer, DefaultVerifier>.Diagnostic(diagnosticId);
```
sharwell, line 206: This ended up getting split into two methods:

```
public static DiagnosticResult Diagnostic()
  => CSharpAnalyzerVerifier<TAnalyzer, DefaultVerifier>.Diagnostic();

public static DiagnosticResult Diagnostic(string diagnosticId)
  => CSharpAnalyzerVerifier<TAnalyzer, DefaultVerifier>.Diagnostic(diagnosticId);
```

sharwell:
The main issue I have is the discrepancy between the README and the actual code we produce with the new analyzer/code fix project template. The true code is listed here:

https://github.com/dotnet/roslyn-sdk/tree/main/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Test/Verifiers

MattKotsenas:
> The main issue I have is the discrepancy between the README and the actual code we produce with the new analyzer/code fix project template. The true code is split across several files here:
> 
> https://github.com/dotnet/roslyn-sdk/tree/main/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Test/Verifiers

The code we're currently pointing people to doesn't compile, so I'm happy to make any requested changes, but I don't think the status quo is sufficient here.

I'll push an update with the requested change to the `Diagnostic` method, and then I see two main paths forward:

1. Leave the README as a simple example people can use as a jumping off point
2. Remove the same and instead point the directory you linked

Let me know what you think, or if there are other solutions I haven't considered. Thanks!

MattKotsenas:
OK, replaced the snippet entirely with a link to the template and a link to instructions on instantiating the template using VS instead.

I don't _love_ this solution because it's another layer of indirection and relies on devs either parsing the template by hand or using Visual Studio, but it addresses the potential for drift between docs.

@sharwell, mind taking another look? Thanks!

sharwell (approved):

PR #1160, opened 2024-06-07
Update Testing README to add best practice for caching ReferenceAssemblies

MattKotsenas:
Add a "best practice" and example of caching `ReferenceAssemblies` between tests. New users of the testing infrastructure often set the reference assemblies in a constructor, but don't use `static` or another method to share the instance between tests resulting in a lot of re-resolution and slow tests. As best I can tell this guidance isn't documented anywhere, but all the tests seem to do it.

PR #1176, opened 2024-08-13
Implement a cache for ReferenceAssemblies instances

sharwell:
Avoids the need to manually keep track of identical instances across test suites.

Supersedes #1160

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
@@ -25,7 +25,7 @@ using NuGet.Packaging.Signing;
 
 namespace Microsoft.CodeAnalysis.Testing
 {
-    public sealed partial class ReferenceAssemblies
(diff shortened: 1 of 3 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -123,14 +125,83 @@ public static ReferenceAssemblies Default
...
+            return hashCode;
+#endif
+        }
+
+        public override bool Equals(object? obj)
+            => Equals(obj as ReferenceAssemblies);
+
+        public bool Equals(ReferenceAssemblies? other)
+        {
+            return other is not null
```
ToddGrun, line 185: >return other is not null [](http://example.com/codeflow?start=12&length=24)

maybe a reference equality check preceeding all the other work?

ToddGrun: oh, it looks like the GetOrAddReferenceAssemblies call takes in a newly constructed one, so wouldn't really have any benefit

ToddGrun (approved):
:shipit:

PR #1179, opened 2024-08-21
Improve exception message for VerifyCodeFixAsync

### LINKED ISSUE #697: Improve exception message when VerifyCodeFixAsync(code, code) fails
See https://github.com/dotnet/roslyn-analyzers/pull/4700/files#r558434203

MattFromRVA:
Fixes #697

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
@@ -692,11 +692,11 @@ namespace Microsoft.CodeAnalysis.Testing
 
             try
             {
-                if (expectedNumberOfIterations == 0 && numberOfIterations <= 0)
+                if (expectedNumberOfIterations == 0)
                 {
                     verifier.Equal(expectedNumberOfIterations, expectedNumberOfIterations - numberOfIterations, "No code fixes were expected, but a fix was offered that made no changes to the code.");
                 }
-                else if (expectedNumberOfIterations >= 0)
+                else if (expectedNumberOfIterations > 0)
                 {
                     verifier.Equal(expectedNumberOfIterations, expectedNumberOfIterations - numberOfIterations, $"Expected '{expectedNumberOfIterations}' iterations but found '{expectedNumberOfIterations - numberOfIterations}' iterations.");
                 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
```diff
@@ -692,7 +692,11 @@ private async Task VerifyProjectAsync(ProjectState newState, Project project, IV
 
             try
             {
-                if (expectedNumberOfIterations >= 0)
+                if (expectedNumberOfIterations == 0 && numberOfIterations <= 0)
```
sharwell, line 695: 💭 We don't even need to check `numberOfIterations` right?

```suggestion
                if (expectedNumberOfIterations == 0)
```

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
```diff
@@ -692,7 +692,11 @@ private async Task VerifyProjectAsync(ProjectState newState, Project project, IV
...
+                {
+                    verifier.Equal(expectedNumberOfIterations, expectedNumberOfIterations - numberOfIterations, "No code fixes were expected, but a fix was offered that made no changes to the code.");
+                }
+                else if (expectedNumberOfIterations >= 0)
```
sharwell, line 699: If the previous change is taken, this can also be simplified:

```suggestion
                else if (expectedNumberOfIterations > 0)
```

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
```diff
@@ -862,7 +866,11 @@ private async Task VerifyProjectAsync(ProjectState newState, Project project, IV
 
             try
             {
-                if (expectedNumberOfIterations >= 0)
+                if (expectedNumberOfIterations == 0 && numberOfIterations <= 0)
```
sharwell, line 869: Same as above

```suggestion
                if (expectedNumberOfIterations == 0)
```

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.CodeFix.Testing/CodeFixTest`1.cs
```diff
@@ -862,7 +866,11 @@ private async Task VerifyProjectAsync(ProjectState newState, Project project, IV
...
+                {
+                    verifier.Equal(expectedNumberOfIterations, expectedNumberOfIterations - numberOfIterations, "No code fixes were expected, but a fix was offered that made no changes to the code.");
+                }
+                else if (expectedNumberOfIterations >= 0)
```
sharwell, line 873: Same as above

```suggestion
                else if (expectedNumberOfIterations > 0)
```

sharwell (approved):

PR #1181, opened 2024-08-22
Add validation for additional unnecessary locations

sharwell:


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -558,16 +560,43 @@ private void VerifyDiagnosticResults(IEnumerable<(Project project, Diagnostic di
                 else
                 {
                     VerifyDiagnosticLocation(analyzers, actual.diagnostic, expected, actual.diagnostic.Location, expected.Spans[0], verifier);
+                    int[] unnecessaryIndices = { };
+                    if (actual.diagnostic.Properties.TryGetValue(WellKnownDiagnosticTags.Unnecessary, out var encodedUnnecessaryLocations))
+                    {
+                        verifier.True(actual.diagnostic.Descriptor.CustomTags.Contains(WellKnownDiagnosticTags.Unnecessary), "Diagnostic reported extended unnecessary locations, but the descriptor is not marked as unnecessary code.");
```
ToddGrun, line 566: >actual.diagnostic.Descriptor.CustomTags.Contains(WellKnownDiagnosticTags.Unnecessary) [](http://example.com/codeflow?start=38&length=85)

Could this be verified as false in WellKnownDiagnosticTags.Unnecessary isn't in the Properties?

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -558,16 +560,43 @@ private void VerifyDiagnosticResults(IEnumerable<(Project project, Diagnostic di
...
+                        var match = EncodedIndicesSyntax.Match(encodedUnnecessaryLocations);
+                        verifier.True(match.Success, $"Expected encoded unnecessary locations to be a valid JSON array of non-negative integers: {encodedUnnecessaryLocations}");
+                        unnecessaryIndices = match.Groups["Index"].Captures.OfType<Capture>().Select(capture => int.Parse(capture.Value)).ToArray();
+                        verifier.NotEmpty(nameof(unnecessaryIndices), unnecessaryIndices);
```
ToddGrun, line 570: >verifier.NotEmpty(nameof(unnecessaryIndices), unnecessaryIndices); [](http://example.com/codeflow?start=24&length=66)

nit: not sure which I like better, just wanted to point out that the regex could have been written to not accept an empty number of indices

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -855,10 +884,23 @@ private static string FormatDiagnostics(ImmutableArray<DiagnosticAnalyzer> analy
                 }
                 else
                 {
-                    AppendLocation(diagnostics[i].Location);
-                    foreach (var additionalLocation in diagnostics[i].AdditionalLocations)
+                    // The unnecessary code designator is ignored for the primary diagnostic location.
+                    AppendLocation(diagnostics[i].Location, isUnnecessary: false);
```
ToddGrun, line 888: >diagnostics[i] [](http://example.com/codeflow?start=35&length=14)

nit: feels like this code would be a bit simpler if diagnostics[i] was just in a local

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
```diff
@@ -855,10 +884,23 @@ private static string FormatDiagnostics(ImmutableArray<DiagnosticAnalyzer> analy
...
+
+                    int[] unnecessaryIndices = { };
+                    if (diagnostics[i].Properties.TryGetValue(WellKnownDiagnosticTags.Unnecessary, out var encodedUnnecessaryLocations))
+                    {
+                        var match = EncodedIndicesSyntax.Match(encodedUnnecessaryLocations);
+                        if (match.Success)
+                        {
+                            unnecessaryIndices = match.Groups["Index"].Captures.OfType<Capture>().Select(capture => int.Parse(capture.Value)).ToArray();
```
ToddGrun, line 896: >ToArray [](http://example.com/codeflow?start=142&length=7)

perf doesn't matter here, but this is used more as a set than an array.

tests/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Testing.Utilities/TestAnalyzers/AbstractHighlightTokensAnalyzer.cs
```diff
@@ -10,8 +10,13 @@ namespace Microsoft.CodeAnalysis.Testing.TestAnalyzers
     public abstract class AbstractHighlightTokensAnalyzer : DiagnosticAnalyzer
     {
         protected AbstractHighlightTokensAnalyzer(string id, params int[] tokenKinds)
+            : this(id, customTags: new string[0], tokenKinds)
```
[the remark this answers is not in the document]
sharwell: ➡️ Not available in all target frameworks.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: ➡️ No, a primary use case for faded diagnostics is when the primary diagnostic location is faded. There would be no additional property in this case.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: ➡️ I felt it would be harder to understand why the regex failed

ToddGrun (approved):
:shipit:

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: ➡️ Will keep this form, since the collection is typically small, and if we ever want to add assertions about uniqueness or order it will be clearer what we are starting with.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/AnalyzerTest`1.cs
sharwell: ➡️ Simplified

PR #1186, opened 2024-08-22
WithArguments call validation

### LINKED ISSUE #516: Fail when call to WithArguments provides more arguments that the resource/message expects
Currently when your message has X format items and you call `WithArguments` with more than X arguments no issue is raised if the first X arguments match the expected. It would be nice if it was failing for extra parameters this way we could avoid regressions when the message suddenly uses less arguments.

From https://github.com/dotnet/roslyn-analyzers/pull/3483#discussion_r415911206

MattFromRVA:
Fixes #516

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/DiagnosticResult.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/DiagnosticResult.cs
@@ -224,6 +224,17 @@ namespace Microsoft.CodeAnalysis.Testing
 
         public DiagnosticResult WithArguments(params object[] arguments)
         {
+            if (MessageFormat != null)
+            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/DiagnosticResult.cs
```diff
@@ -224,6 +224,17 @@ public DiagnosticResult WithOptions(DiagnosticOptions options)
 
         public DiagnosticResult WithArguments(params object[] arguments)
         {
+            if (MessageFormat != null)
```
sharwell, line 227: 💭 What about moving this validation to the `Message` property? Currently the validation will not work for these cases:

1. False exception: `WithArguments` is called before `WithMessageFormat`
2. False non-exception: `WithMessageFormat` is called with 3 placeholders, then `WithArguments` is called with 3 arguments, then `WithMessageFormat` is called again with 2 placeholders.

dibarbet:

PR #1196, opened 2024-12-01
Build for .NET 5, 6, 7, 8, 9 & 10 and ReferenceAssemblies.Default to locate each respective runtime

thomhurst:
## Summary

- Add `net5.0` through `net10.0` to `TestingLibraryTargetFrameworks` and `TestingLibrarySourceGeneratorTargetFrameworks`, enabling the testing libraries to be built for each modern .NET runtime
- Update `ReferenceAssemblies.Default` to return the correct reference assemblies for the target framework being compiled against (e.g. `Net.Net100` on .NET 10, `Net.Net90` on .NET 9, etc.) via conditional compilation
- Update `global.json` to .NET 10 SDK to support building `net10.0` targets
- Add `System.Resources.Extensions` package reference to fix VSSDK1012 build error with newer SDK versions
- Add `CheckNotRecommendedTargetFramework=false` to allow building .NET Standard 1.6 with the .NET 10 SDK

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Directory.Build.props
+++ b/src/Microsoft.CodeAnalysis.Testing/Directory.Build.props
@@ -10,17 +10,16 @@
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
 
   <PropertyGroup>
-    <!--
-      netcoreapp3.1: Ensures full support for nullable reference types
-      netstandard2.0/net472: Roslyn 3.x
-    -->
-    <TestingLibraryTargetFrameworks>netcoreapp3.1;netstandard2.0;net461;net472</TestingLibraryTargetFrameworks>
+    <!-- Required for the .NET 10 SDK to build out-of-support target frameworks -->
+    <CheckNotRecommendedTargetFramework>false</CheckNotRecommendedTargetFramework>
+
+    <TestingLibraryTargetFrameworks>net10.0;net9.0;net8.0;net7.0;net6.0;net5.0;netcoreapp3.1;netstandard1.6;netstandard2.0;net452;net46;net461;net472</TestingLibraryTargetFrameworks>
 
     <!--
       netcoreapp3.1: Ensures full support for nullable reference types
       netstandard2.0/net472: Roslyn 3.x, starting with 3.8.0
     -->
-    <TestingLibrarySourceGeneratorTargetFrameworks>netcoreapp3.1;netstandard2.0;net472</TestingLibrarySourceGeneratorTargetFrameworks>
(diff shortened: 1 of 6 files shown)


thomhurst:
Got a few errors switching to the .NET 9 SDK which I think I've managed to fix, but I've got a new error now that I have no idea what it means or how to fix it:

> VSSDK(0,0): error VSSDK1012: (NETCORE_ENGINEERING_TELEMETRY=Build) The cto merger could not read the data from 'D:\a\_work\1\s\artifacts\obj\Roslyn.SyntaxVisualizer.Extension\Release\net472\VSPackage.resources.original'. This .resources file should not be read with this reader. The resource reader type is "System.Resources.Extensions.DeserializingResourceReader, System.Resources.Extensions, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51".

AraHaan:
This PR can be updated to .NET 10 as well.

thomhurst:
@AraHaan added

src/Microsoft.CodeAnalysis.Testing/Directory.Build.props
```diff
@@ -10,17 +10,16 @@
...
   <PropertyGroup>
-    <!--
-      netcoreapp3.1: Ensures full support for nullable reference types
-      netstandard2.0/net472: Roslyn 3.x
-    -->
-    <TestingLibraryTargetFrameworks>netcoreapp3.1;netstandard2.0;net461;net472</TestingLibraryTargetFrameworks>
+    <!-- Required for the .NET 10 SDK to build out-of-support target frameworks -->
+    <CheckNotRecommendedTargetFramework>false</CheckNotRecommendedTargetFramework>
+
+    <TestingLibraryTargetFrameworks>net10.0;net9.0;net8.0;net7.0;net6.0;net5.0;netcoreapp3.1;netstandard1.6;netstandard2.0;net452;net46;net461;net472</TestingLibraryTargetFrameworks>
```
Youssef1313, line 16: I'm not a maintainer of this project, but IMO it's not worth it to pack for all these TFMs.

I would probably just add `net10.0` to the existing TFMs. So it's .NET 10, .NET Core 3.1, .NET Standard 2.0, and .NET Framework 4.7.2.

src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
```diff
@@ -88,8 +88,26 @@ public static ReferenceAssemblies Default
...
+                return NetStandard.NetStandard16;
+#elif NETSTANDARD2_0
                 return NetStandard.NetStandard20;
+#elif NET452
+                return NetFramework.Net452.Default;
+#elif NET46
+                return NetFramework.Net46.Default;
 #elif NET461
                 return NetFramework.Net461.Default;
 #elif NET472
```
teo-tsirpanis, line 113: I wonder if it would be better to get the reference assemblies' versions from the running framework's version, instead of adding more APIs and ifdefs every year.

dibarbet:

PR #1213, opened 2025-03-01
Update xcopy-msbuild to latest available version

benhopkinstech:
Updated `xcopy-msbuild` in `global.json` to the latest available version on dotnet-eng

I am raising this as I made the change in [aspnetcore](https://github.com/dotnet/aspnetcore) via https://github.com/dotnet/aspnetcore/pull/60666 and spotted this repository was using an older version of 17.x

### DIFF
--- a/global.json
+++ b/global.json
@@ -14,7 +14,7 @@
     "vs": {
       "version": "17.8.0"
     },
-    "xcopy-msbuild": "17.8.1-2"
+    "xcopy-msbuild": "17.13.0"
   },
   "msbuild-sdks": {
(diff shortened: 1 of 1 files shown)


JoeRobich:
> Could one of you provide a review on this?

@benhopkinstech Since we are pinning to an older SDK in our global.json, it isn't strictly necessary to update our MSBuild. Bumps to the xcopy-msbuild version are necessary based on the [minimumMSBuildVersion](https://github.com/dotnet/sdk/blob/v8.0.101/src/Layout/redist/minimumMSBuildVersion) of the SDK being used.

PR #1245, opened 2026-01-14
Net10: Avoid directly constructing NuGetFrameworkSorter

samblackburn:
Hi,

This is more of a "how should this work?" question than a PR I expect to be merged tbh, so feel free to close since I have a workaround and presumably you have some net10 work planned at some point.

## Why we want to upgrade NuGet.Framework
We're trying to move an analyzer to `net10.0` but our tests get thrown off by version mismatches unless we use `ReferencedAssemblies.Net.Net100` which is only available in `NuGet.Frameworks 7.0.1`.

```
// error CS1705: Assembly [...] uses 'System.Runtime, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'System.Runtime' with identity 'System.Runtime, Version=9.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
DiagnosticResult.CompilerError("CS1705").WithArguments([...], "System.Runtime, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a", "System.Runtime", "System.Runtime, Version=9.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"),
```

## Why upgrading NuGet.Framework breaks our tests
Unfortunately, `ReferencedAssemblies.cs` [calls](https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs#L523) the `NuGetFrameworkSorter` constructor, which has been obsolete for a while and was recently [made private](https://github.com/NuGet/NuGet.Client/pull/6529/changes#diff-32f05f393681e7b72e205e0919d630cd8d853640357db06c7e3e294f8839f266L20-R18):

```
System.MethodAccessException : Attempt by method 'Microsoft.CodeAnalysis.Testing.ReferenceAssemblies+<ResolveCoreAsync>d__55.MoveNext()' to access method 'NuGet.Frameworks.NuGetFrameworkSorter..ctor()' failed.
```

## What this PR is trying to do
I've got into a bit of a mess trying to reference `NuGet.Framework 7.0.1` in this repo - anyone have any suggestions? Should I add another target framework for `net10`?

## Workaround
For now I'm using a workaround - instead of `ReferenceAssemblies.Net.Net90` use:
``` c#
new ReferenceAssemblies(
                "net10.0",
                new PackageIdentity("Microsoft.NETCore.App.Ref", "10.0.2"),
                Path.Combine("ref", "net10.0"));
```

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs
@@ -271,7 +271,7 @@ namespace Microsoft.CodeAnalysis.Testing
         /// <seealso href="https://martinbjorkstrom.com/posts/2018-09-19-revisiting-nuget-client-libraries"/>
         private async Task<ImmutableArray<MetadataReference>> ResolveCoreAsync(string language, CancellationToken cancellationToken)
         {
-            var settings = string.IsNullOrEmpty(NuGetConfigFilePath) ? Settings.LoadDefaultSettings(root: null) : Settings.LoadSpecificSettings(root: null, NuGetConfigFilePath);
+            var settings = string.IsNullOrEmpty(NuGetConfigFilePath) ? Settings.LoadDefaultSettings(root: null) : Settings.LoadSpecificSettings(root: null!, NuGetConfigFilePath!);
             var sourceRepositoryProvider = new SourceRepositoryProvider(new PackageSourceProvider(settings), Repository.Provider.GetCoreV3());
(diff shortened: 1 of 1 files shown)


evgenyfedorov2:
Have the same issue.
Thanks for the workaround - it works for me!

samblackburn:
Closing as it looks like the root cause of the problem was that we had an unnecessary pinned package:
`<PackageReference Include="NuGet.Protocol" Version="7.0.1"/>`
This was to avoid a vulnerability at the time but removing the pin has fixed the issue for us.

PR #1257, opened 2026-05-22
Update NuGet.* versions and add .NET 8 TFM

333fred:
The newest versions of the NuGet.* libraries have a breaking change that impacts consumers. When a consumer updates those packages to a newer version to avoid CVE alerts, the testing SDK breaks. We update the versions we use, address a few nullable warnings introduced, and add an explicit .NET 8 TFM for support of this new version.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest`1.cs
+++ b/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest`1.cs
@@ -368,7 +368,7 @@ namespace Microsoft.CodeAnalysis.Testing
 
                 if (x.content is null || y.content is null)
                 {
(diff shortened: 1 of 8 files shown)


src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/CodeActionTest`1.cs
```diff
@@ -368,7 +368,7 @@ public bool Equals((string filename, SourceText content) x, (string filename, So
 
                 if (x.content is null || y.content is null)
                 {
-                    return ReferenceEquals(x, y);
+                    return ReferenceEquals(x.content, y.content);
```
333fred, line 371: Note: this is technically a behavior change, as the previous code incorrectly boxed x and y, resulting in an always false result.

JoeRobich (approved):

JoeRobich (approved):

PR #1266, opened 2026-07-22
Remove StyleCop analyzers

mwiemer-microsoft:
Works on https://github.com/dotnet/roslyn/issues/84573. dotnet/roslyn intentionally doesn't use StyleCop, so we don't want it in this repo either.


mwiemer-microsoft:
@RikkiGibson requesting review :)

JoeRobich (approved):

